PR #14, opened 2023-02-01
Get ready for automated beta release, enable all C.I. checks

NullVoxPopuli:
## What's included in this repo's C.I.?

 - assert that source maps and type-maps are generated
 - support multiple versions of typescript, per https://www.semver-ts.org/
 - support multiple versions of ember-source, to ensure a smooth upgrade path for downstream consumers
 - support embroider so that downstream consumers can be assured that their dependencies are "embroider strict mode safe"
- test with no lockfile to try to catch unintended ecosystem deviations / breaking changes (such as what may happen with babel / rollup / etc)

(all of the above is standard in ember OSS projects 🎉 )

extra stuff for this repo:
- publish the docs site and post the preview URLs to PRs
- configure the auto-release system to do a series of beta versions until we're ready for 1.0.0

## What is supported?

- minimum ember-source 4.8 -- we can start pretty new, but this also allows us to use the nicest features. Additionally, since we are beta, we can move the minimum ember-source to whichever ember-source ships the built-in-types (as non-preview), without a breaking change
- TS 4.8 and 4.9 -- we cannot support earlier than TS 4.8, because Glint does not support prior to TS 4.8
- embroider-strict and classic builds


.changeset/pre.json
```diff
@@ -0,0 +1,10 @@
+{
+  "mode": "pre",
```
NullVoxPopuli, line 2: this configures our release automation to only do beta releases for now

.github/workflows/ci.yml
```diff
@@ -89,77 +89,141 @@ jobs:
...
+          persist-credentials: false
+      - uses: ./.github/actions/pnpm
+      - uses: ./.github/actions/download-built-package
+      - uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          message: |+
+            ## Preview URLs
+            GH Env: ${{ needs.PublishDocstoCloudflarePages.outputs.env }}
+            docs: ${{ needs.PublishDocstoCloudflarePages.outputs.url }}
+            api docs: ${{ needs.PublishDocstoCloudflarePages.outputs.url }}/api/modules.html
```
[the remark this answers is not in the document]
NullVoxPopuli: we don't yet have api docs, but when we do! this URL will be ready!
(I'm thinking for toucan-core, this makes the most sense for test-helpers and all that)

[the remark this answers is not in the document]
ynotdraw: > (I'm thinking for toucan-core, this makes the most sense for test-helpers and all that)
makes sense to me 👍

NullVoxPopuli:
This PR addresses a real bug in ember-browser-services and unblocks this PR: https://github.com/CrowdStrike/ember-browser-services/pull/364

docs-app/types/glint-registry.d.ts
```diff
@@ -1,8 +1,8 @@
 import '@glint/environment-ember-loose';
 import '@glint/environment-ember-template-imports';
-// Types from libraries
-import '@crowdstrike/ember-oss-docs/glint';
```
NullVoxPopuli, line 4: side-effecting imports are no longer recommended for merging the glint registry, because the number of copies of `@glint/environment-ember-loose/registry` in a project's `node_modules` is too variable.

package.json
```diff
@@ -24,10 +25,15 @@
...
   "devDependencies": {
+    "@changesets/changelog-github": "^0.4.8",
+    "@changesets/cli": "^2.26.0",
     "concurrently": "^7.6.0",
-    "prettier": "^2.8.2"
+    "prettier": "^2.8.3"
   },
   "pnpm": {
+    "overrides": {
+      "@types/eslint": "^8.0.0"
```
NullVoxPopuli, line 35: ember-template-lint's dependency of `@lint-todo` brings in `@types/eslint` v7 🙈

test-app/package.json
```diff
@@ -76,6 +76,7 @@
     "broccoli-asset-rev": "^3.0.0",
     "concurrently": "^7.6.0",
     "ember-auto-import": "^2.5.0",
+    "ember-browser-services": "^4.0.4",
```
NullVoxPopuli, line 79: this works around an issue in embroider which will be fixed in the next release -- ultimately though, I think we may want to make this a peer _eventually_ anyway, depending on what our components end up doing -- only time will tell though.

ynotdraw (approved):
This is exciting!

.github/workflows/ci.yml
```diff
@@ -89,77 +89,141 @@ jobs:
...
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/pnpm
+      - name: Install Dependencies (without lockfile)
+        run: rm pnpm-lock.yaml && pnpm install
+      - uses: ./.github/actions/download-built-package
+      - run: pnpm --filter test-app test:ember
```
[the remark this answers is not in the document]
NullVoxPopuli: yeah, either that, or _accidental_ breaking changes, such as what happened recently here: https://github.com/wessberg/rollup-plugin-ts/issues/202

Ultimately SemVer is a social contract, and as such, is prone to human error 😅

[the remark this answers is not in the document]
ynotdraw: Totally! Makes sense to me. Thank ya for the context

nicolechung:
>docs: https://0cc0132f.ember-toucan-core.pages.dev/

Woohoo! Docs!

ember-toucan-core/package.json
```diff
@@ -85,7 +85,7 @@
     "concurrently": "^7.2.1",
     "ember-cli-htmlbars": "^6.1.1",
     "ember-source": "~4.9.2",
-    "ember-template-lint": "^4.0.0",
+    "ember-template-lint": "^5.3.3",
```
nicolechung, line 88: Wow...so many packages updated in just a month

.github/workflows/ci.yml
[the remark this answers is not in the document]
nicolechung: Explains a lot to me too!

nicolechung (approved):
🎉

.changeset/pre.json
simonihmig: How exactly does that work here? Will changeset pick a version for us? And if so, will it be a beta release like `1.0.0-beta.0`, or will it be 0.x prereleases like `0.1.0`?

.changeset/pre.json
```diff
@@ -0,0 +1,10 @@
...
+  "tag": "beta",
+  "initialVersions": {
+    "docs-app": "0.0.0",
+    "@crowdstrike/ember-toucan-core": "0.0.0",
```
simonihmig, line 6: This depends on the answer to the previous question, but just to point this out: the `0.0.x` releases are a bit tricky and awkward IMO, as according to how npm version modifiers work any new patch release like `0.0.x+1` is considered a breaking change, as `^0.0.1` would _not_ allow `0.0.2` to be used. So unless our API is really like super unstable, I would start with `0.1.0` (as `^0.1.0` would cover all `0.1.x` releases then).

.github/workflows/ci.yml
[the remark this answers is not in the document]
simonihmig: accidental breaking changes also includes bugs, which also tend to not follow semver 🤪

Even when that's not our fault, it's good to have visibility about things failing, to raise issues upstream and prevent "works on my machine"™ like scenarios...

.github/workflows/ci.yml
```diff
@@ -89,77 +89,141 @@ jobs:
...
+    name: Post Preview URL as comment to PR
+    runs-on: ubuntu-latest
+    needs: PublishDocstoCloudflarePages
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/pnpm
+      - uses: ./.github/actions/download-built-package
+      - uses: marocchino/sticky-pull-request-comment@v2
```
simonihmig, line 219: I have used similar things before, but this one is really nice with its update capabilities! 😍

.github/workflows/ci.yml
```diff
@@ -89,77 +89,141 @@ jobs:
...
+          persist-credentials: false
+      - uses: ./.github/actions/pnpm
+      - uses: ./.github/actions/download-built-package
+      - run: pnpm build:docs-app
+      - name: Publish to Cloudflare Pages
+        id: publishStep
+        uses: cloudflare/pages-action@1
+        with:
+          apiToken: ${{ secrets.UX_OSS_CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.UX_OSS_CLOUDFLARE_ACCOUNT_ID }}
```
simonihmig, line 202: I don't have the powers to see the repo/org secrets in this Github, so wondering, are these repo- or org-wide secrets? Or to rephrase: is this already available e.g. in the forms repo?

.github/workflows/ci.yml
```diff
@@ -89,77 +89,141 @@ jobs:
...
+  release:
+    name: Release
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    needs:
+      - default_tests
+      - floating_tests
+      - typecheck
+      - try_scenarios
```
simonihmig, line 145: I guess we cannot split things up a bit into different workflows (like CI/release/docs) due to not being able to reference `needs` across workflows, right?

docs-app/types/glint-registry.d.ts
```diff
@@ -17,7 +17,7 @@ declare module '@ember/modifier' {
 }
 
 declare module '@glint/environment-ember-loose/registry' {
-  export default interface Registry {
+  export default interface Registry extends OSSDocs {
```
simonihmig, line 20: :+1:

ember-toucan-core/package.json
```diff
@@ -96,7 +96,7 @@
     "postcss": "^8.2.14",
     "prettier": "^2.8.3",
     "prettier-plugin-ember-template-tag": "^0.3.0",
-    "rollup": "^2.67.0",
+    "rollup": "^3.12.1",
```
simonihmig, line 99: Oh! Yeah, I remember I wanted to update this also in the blueprint, but didn't look yet into what it takes (breaking changes)... seems... nothing?

simonihmig (commented):
Nice! 🚀

.changeset/pre.json
NullVoxPopuli: not sure when things are pre 1.0 -- I assume it'll do 1.0.0-beta.x until we turn off this mode. It'll also update the `beta` tag on npm -- I did a test with this on ember-toucan-styles, but that was for a 2.0.0 beta that I ended up abandoning

.github/workflows/ci.yml
NullVoxPopuli: correct

.github/workflows/ci.yml
NullVoxPopuli: they are org-wide --  tokens configured in the org settings -- and are explicitly allowed per repo that needs them (as opposed to all repos by default) -- if forms is ready for docs, I can add these to the forms repo 🎉

Issue #10, opened 2023-01-19
Create field component


nicolechung:
Create a field component.

This will be mostly styling. 

Field is kind of a form-element wrapper exposing label, errors, etc but is not opinionated on the "control" it renders (e.g., an input vs textarea vs select)

Rough API:
```hbs
<!-- Custom input component -->
<form.field as |field|>
    <ToucanCoreField @name="lastName" @label="Last Name">
        <:control>
            <form.controls.Input
                {{on 'change' this.doWhatever}} 
                required 
            />
        </:control>
    </ToucanCoreField>
</form.field>
```

ynotdraw:
Thinking out loud:

I'm kind of thinking we build out _only_ the primitive controls + Field at this point.  Essentially worry about the **presentational** pieces at the moment and "ignore" the `<Form` API for the time being.  I feel like Toucan Core should focus on providing presentational building blocks to consumers, and it's up to them to build the lego of their dreams using our blocks.  By providing only the presentational pieces and allowing consumers to wire things up, it opens the door a bit as to how they can integrate their favorite packages.

We may use `ember-headless-form` for example, but someone else may want to use some other form library or even write their own!  Instead of tying our components to a single form solution, I think pushing it to consumers makes for better composition and separation of concerns (e.g., presentational vs presentation+functional).

---

To support above, here's what we could maybe do.  I'm not sure about file paths / component naming, I also took this from I think either you @NullVoxPopuli or someone else wrote up:

```
<Field @name="..">  
 <:label></:label>
 <:hint></:hint>  
 <:control></:control>  
 <:inlayButtons></:inlayButtons>
 <:error></:error>  
</Field>  
```

Field is essentially a layout component.  It yields back named blocks to put content in particular sections.   The field implementation may be something like ( _super rough_ idea - can land on something better if this seems like a path we want to take ):

```html
// src/components/field.hbs

<div ...attributes>
  <label class="classes">{{yield to="label"}}</label>
  <p class="classes" aria-describedby="">{{yield to="hint"}}</p>
  <div class="classes relative">
    {{yield to="control"}}
    <div class="classes absolute right-0 inset-y-0">{{yield to="inlayButtons"}}</div>
  </div>
  <p class="classes" aria-describedby="">{{yield to="error"}}</p>
</div>
````

From a top-down level, this matches how most folks layer forms structurally:

```
<form label>
<some help text>
<the control>
<error messages (sometimes help text is down here too!>
```

---

[...]

PR #27, opened 2023-02-10
feat: Added form field component

nicolechung:
## 🚀 Description
This PR introduces a `Field` component.  It will enable us to build other form controls by providing styled form-elements.

---

## 🔬 How to Test

**CI**
CI (Actions) should all be green 🟢 .  CI tests linting and tests!

**Docs App**

View the docs at https://87acbda6.ember-toucan-core.pages.dev/

**- OR -**

Pull down this repo and run the following:

```bash
pnpm i
cd docs-app
pnpm start
```

View the Field documentation

---

Screenshot 2023-02-14 at 10 06 15 AM


ember-toucan-core/src/components/form/field.ts
```diff
@@ -0,0 +1,20 @@
+import Component from '@glimmer/component';
+
+import Wrapper from './wrapper';
+
+interface ToucanFormFieldComponentSignature {
+  Element: HTMLDivElement;
+  Args: {};
+  Blocks: {
+    // TODO - type these properly!
```
nicolechung, line 9: We tried this:
```
interface ToucanFormWrapperComponentSignature {
  Element: HTMLDivElement;
  Args: {};
  Blocks: {
    default?: [
      {
        // (property) Label: typeof Label
        // Type 'abstract new () => PartiallyAppliedComponent<EmptyObject, {}, AcceptsBlocks<FlattenBlockParams<{ default: { Params: { Positional: []; }; }; }>, HTMLLabelElement>>' is not assignable to type 'typeof ToucanFormLabel'.
        // Cannot assign an abstract constructor type to a non-abstract constructor type.glint:ts(2322)
        Label: typeof Label;
        Hint: typeof Hint;
        Control: typeof Control;
        Error: typeof Error;
      }
    ];
// Error: Property '[Data]' is missing in type 'abstract new () => PartiallyAppliedComponent<EmptyObject, {}, AcceptsBlocks<FlattenBlockParams<{ default: { Params: { Positional: []; }; }; }>, HTMLDivElement>>' but required in type 'TemplateOnlyComponent<ToucanFormHintComponentSignature>'
    hint?: [component: typeof Hint];
    control?: [component: typeof Control];
    label?: [component: typeof Label];
    error?: [component: typeof Error];
  };
}
```

We were also getting an error

ember-toucan-core/src/components/form/control.ts
```diff
@@ -0,0 +1,11 @@
+import templateOnlyComponent from '@ember/component/template-only';
+
+export interface ToucanFormControlComponentSignature {
+  Element: HTMLElement;
```
NullVoxPopuli, line 4: there is no element, should be `null`

ynotdraw: Thank you! Fixed!

ember-toucan-core/src/components/form/field.ts
nicolechung: If anyone knows how to fix those errors (and also: what's wrong), we're looking for advice!

ember-toucan-core/src/components/form/wrapper.hbs
```diff
@@ -0,0 +1,15 @@
...
+      Hint=(component this.Hint)
+      Control=(component this.Control)
+      Error=(component this.Error)
+    )
+  }}
+{{else}}
+  {{yield (component this.Label) to="label"}}
+  {{yield (component this.Hint) to="hint"}}
+  {{yield (component this.Control) to="control"}}
+  {{yield (component this.Error) to="error"}}
```
ynotdraw, line 14: One of the things @nicolechung and I talked about, but I'm not personally _quite_ sure provides a ton of value is yielding back named blocks here versus only doing the contextual components.

So if you do:

```hbs
<Field>
  <:label>Label<:/label>
</Field>
```

You'll still get the hint + error divs rendered like https://github.com/CrowdStrike/ember-toucan-core/issues/10#issuecomment-1422872713 mentions, so you'll get extra stuff in the DOM that you don't (potentially) care about.  And in the case of the Error, you'll have a lingering error icon.

I _feel_ like the contextual component path is a better API for consumers, as folks can add only what they need like:


```hbs
// Taken from https://github.com/CrowdStrike/ember-toucan-core/compare/10-field-prototype-w-nicole#diff-6b74531eede9b2af7db5be79e6cc466ce08e9f1c7c604348a819b3a0825f8d76
<template>
  <Field as |field|>
    <field.Label>{{@label}}</field.Label>

    {{#if @helperText}}
      <field.Hint>{{@helperText}}</field.Hint>
    {{/if}}

    <field.Control>
      <input class="border-critical bg-blue" ...attributes />
    </field.Control>

    {{#if @error}}
      <field.Error>{{@error}}</field.Error>
    {{/if}}
  </Field>
</template>
```

and then usage of above:


```hbs
<InputField @label="Input" />
<InputField @label="Input" @helperText="help!" />
<InputField @label="Input" @helperText="help!" @error="error" />
```

---

I also don't have strong opinions, I think it's just more my brain trying to figure out a use case for the named blocks if they _can't_ conditionally render blocks they don't care about.  Any thoughts around this or guidance?

ember-toucan-core/src/components/form/field.ts
NullVoxPopuli: I think headless-form has examples of this: https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/components/headless-form.ts#L50

nicolechung: So we went and tried that, also following this guide: https://typed-ember.gitbook.io/glint/using-glint/ember/contextual-components.

But currently stuff like `Label` and `Hint` do not have any `args` so the guide recommends the approach above (just using `typeof Label`. Is it possible to book some time next week to pair on this?

docs/components/field/demo/base-demo.md
```diff
@@ -0,0 +1,19 @@
+```hbs template
+<Form::Field as |field|>
+  <field.Label>First name</field.Label>
```
simonihmig, line 3: In headless-form, I yielded components using a lowercase name, for no particular reason. Should we try to establish a convention (within CS at least)?

I don't have any strong feelings for one or the other, has anyone of you?

docs/components/field/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+  {{#if @error}}
+    <field.Error>{{@error}}</field.Error>
+  {{/if}}
+</Form::Field>
+```
+
+### Extending the Components
+
+We have `...attributes` added to each sub component of the Field, so HTML attributes Ember modifiers can be added.
```
simonihmig, line 36:
```suggestion
We have `...attributes` added to each sub component of the Field, so HTML attributes and Ember modifiers can be added.
```

ember-toucan-core/src/components/form/label.hbs
```diff
@@ -0,0 +1,3 @@
+<label class="type-md-tight text-body-and-labels block" ...attributes>
```
simonihmig, line 1: headless-form automatically applies `for` to labels, and a matching `id` to the control element, at least when using the provided components. This is the most basic a11y rule we need to follow for forms. So I wonder if/how this component should be concerned about...

The user brings their own control component, so we cannot set an `id` for them. So we could either generate an id (e.g. using `unique-id` helper), automatically set it as `for` on the label, and yield it, so users cann apply it to their controls, like this:

```hbs
<Form::Field as |field|>
  <field.Label>First name</field.Label>
  <field.Control>
    <input id={{field.id}} ...attributes />
  </field.Control>
  <field.Hint>Hint text below the input</field.Hint>
</Form::Field>
```

Or we choose to not make this our concern, but then we should at least call this out in the docs!

For the more opinionated toucan-forms, I think we will wire this all up correctly for the user, right?

ember-toucan-core/src/components/form/wrapper.hbs
```diff
@@ -0,0 +1,15 @@
+{{#if (has-block)}}
+  {{yield
+    (hash
+      Label=(component this.Label)
```
simonihmig, line 4: I think this will fail for Embroider with optimized settings. Basically you are only allowed to use `component` with static strings, or wrapping the dynamic component reference (`this.Label` could be arbitrary things, that Embroider's static analysis is not able to understand) with `ensure-safe-component`. I had the use the latter in headless...

_But_, I don't think we need `component` at all here, when not passing any curried arguments. Just `Label=this.Label` should also work, not?

ember-toucan-core/src/components/form/wrapper.ts
```diff
@@ -0,0 +1,38 @@
...
+import Label from './label';
+
+interface ToucanFormWrapperComponentSignature {
+  Element: HTMLDivElement;
+  Args: {};
+  Blocks: {
+    default?: [
+      {
+        // TODO: Figure out how to resolve these type errors
+        // (property) Label: typeof Label
```
simonihmig, line 15: Maybe not using `component` helper will "fix" this (see other comment)?

ember-toucan-core/src/components/form/wrapper.hbs
simonihmig: For conditionally rendering named blocks, wouldn't something like this work?

```hbs
{{#if (has-block "label")}}
  {{yield to="label"}}
{{/if}}
```

But this is not necessarily better than yielded components...

The way I would think about this is who is in charge of the overall markup and layout? 

If we have strong opinions/guidelines/rules for where to put which element, or having markup that wraps everything, that shouldn't be user-controllable, then named blocks are the better fit. They are more restrictive, or let you be in control, which I think is good for design systems, as that's what they do: they impose restrictions on the what and how.

On the other side, if you want to be less opinionated, and give the user freedom to compose things in arbitrary ways, then yielded components are good. headless-form is a good example of this latter case, it's by its own definition un-opinionated. 

So not sure how the field components fits in here. Some questions, which also might help to understand how opinionated this component needs to be:
* what's the deal with the "hint"? Is it really the case that the user can put them either before or after the control, or doesn't the design system dictate where the should be positioned, like ? (unfortunately, I wasn't able to take part in the latest sessions. Wasn't there cases where the hint was above the control, aligned to the right?)
* don't we need to wrap the whole field in some markup, like tailwind flex classes or something, to position things properly according to our Figma designs? Or is this not a concern of this component?

ember-toucan-core/src/components/form/label.hbs
ynotdraw: Definitely! Thank you for bringing this up as a whole.  Yeah, I think we'll need to figure this out.  @nicolechung and I were discussing this last week as we were pairing.  We'll need to do the `id` + `for`, as well as "link" the hint + error blocks as well via something like `aria-describedby` or similar.  We're still discussing if we do this work as part of _this_ PR or a follow up one.  Our goal at the moment is pairing to get this far enough along that we can work on separate tasks without stepping on each other's toes.  We're getting close!

> The user brings their own control component, so we cannot set an id for them. So we could either generate an id (e.g. using unique-id helper), automatically set it as for on the label, and yield it, so users cann apply it to their controls, like this

Yeah!  I like this idea a lot 😄.  Commit coming soon!

> Or we choose to not make this our concern, but then we should at least call this out in the docs!

Absolutely! If we decide it's not our concern here, we can call it out in the docs.  I'd definitely _prefer_ if we can do as much as we can though! (but maybe separate PR for wiring things up properly?)

> For the more opinionated toucan-forms, I think we will wire this all up correctly for the user, right?

Oh yeah, definitely!

ember-toucan-core/src/components/form/wrapper.hbs
ynotdraw: So I gave the following a shot, which made the contextual component interface happy; however, the named blocks are still throwing glint errors.  I went ahead and added a comment below.

```hbs
{{#if (has-block)}}
  {{yield
    (hash Label=this.Label Hint=this.Hint Control=this.Control Error=this.Error)
  }}
{{else}}
  {{yield this.Label to="label"}}
  {{yield this.Hint to="hint"}}
  {{yield this.Control to="control"}}
  {{yield this.Error to="error"}}
{{/if}}
```

```js
interface ToucanFormWrapperComponentSignature {
  Element: HTMLDivElement;
  Args: {};
  Blocks: {
    // Happy!
    default?: [
      {
        Label: typeof Label;
        Hint: typeof Hint;
        Control: typeof Control;
        Error: typeof Error;
      }
    ];
   // Not happy :( 
    hint?: any;
    control?: any;
    // (property) ToucanFormWrapperComponent.Hint: TemplateOnlyComponent<ToucanFormHintComponentSignature>
    // Argument of type '[TemplateOnlyComponent<ToucanFormHintComponentSignature>]' is not assignable to parameter of type 'TemplateOnlyComponent<ToucanFormLabelComponentSignature>'.
    // Property '[Data]' is missing in type '[TemplateOnlyComponent<ToucanFormHintComponentSignature>]' but required in type 'TemplateOnlyComponent<ToucanFormLabelComponentSignature>'.glint:ts(2345)
    label?: typeof Label;
    error?: any;
  };
}
```

ember-toucan-core/src/components/form/field.ts
NullVoxPopuli: yeah, feel free to throw something on my calendar 🎉

ynotdraw: @nicolechung I can loop you in tomorrow, but after some discussion we got this fixed.  We ended up talking a bit more and decided to allow for named-blocks from components like `Input`, but `Field` being contextual may be a bit better for cases like Checkbox (since it breaks out of our normal vertical-flow).

docs/components/field/index.md
ynotdraw: Thank you!

ember-toucan-core/src/components/form/wrapper.hbs
ynotdraw: Yeah! Ended up doing: `Label=this.Label` instead and it worked like a charm! Thank you for the suggestion.  Also removed the contextual components from `<Field`

ember-toucan-core/src/components/form/label.hbs
ynotdraw: Okay, yeah, after playing around a bit tonight it's a bit tricky as we don't have access directly to the control the user renders.  Due to that, I think your example above is the best we can do.

Here's what we can provide to make things fully accessible, so that the hint + error blocks get read to screenreaders:

```hbs
{{#let (unique-id) as |uniqueId|}}
  {{#let (concat "hint-" uniqueId) as |hintId|}}
    {{#let (concat "error-" uniqueId) as |errorId|}}
      {{yield
        (hash
          Label=this.Label
          Hint=this.Hint
          Control=this.Control
          Error=this.Error
          id=uniqueId
          hintId=hintId
          errorId=errorId
        )
      }}
    {{/let}}
  {{/let}}
{{/let}}
```

And then usage:

```hbs
<Form::Field as |field|>
  <field.Label for={{field.id}}>First name</field.Label>
  <field.Hint id={{field.hintId}}>What should we call you?</field.Hint>
  <field.Control>
    <input
      class='border bg-basement text-titles-and-attributes'
      id={{field.id}}
      aria-describedby='{{field.hintId}} {{field.errorId}}'
      ...attributes
    />
  </field.Control>
  <field.Error id={{field.errorId}}>error</field.Error>
</Form::Field>
```

Since we can't access the control element directly, as that's provided by the consumer, we _can_ provide IDs for consumers to wire things up as they please.  Then they can put the `for` + `id` + `aria-describedby` in the correct places.  What's your feeling about something like this?

simonihmig: Yeah, seems good overall.

Some (minor) notes:

* we could wire up the `for` attribute on the label already, as we do have control over the `<label>` element. headless-form does this. But not sure if that's really preferable here. We have two sides of the same coin here, `for` on the label, and `id` on the control. But auto-applying `for` would obfuscate the one, will leaving the other for the user to do. So maybe letting the user do both is less confusing? I'm 50/50 on this one, so your call! :wink:
* FWIW, I used `aria-errormessage` for linking from the control to the element holding the errors., instead of `aria-describedby` with multiple IDs. Not sure which is better, or if there is any significant difference. But "errormessage" sounded more specific, so I chose that one...
* you don't have to use those nested `{{#let}}`, you can also introduce multiple new local vars in a single one, like [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/components/-private/field.hbs#L1-L7)

ember-toucan-core/src/components/form/wrapper.hbs
simonihmig: Sounds great! Thanks for the thorough explanation, especially related to the "what is the concern of this" question!

> I'm curious if you think there's too much overlap between this and what ember-headless-form provides 

Yeah, I am seeing that a bit at least. But I guess we cannot prevent this, when Field should also be usable standalone, without a full-fledged form. I also think this is not terrible.

ember-toucan-core/src/components/form/label.hbs
ynotdraw: Appreciate all of the notes! This is very helpful.   I like the idea of using `aria-errormessage` + `aria-invalid="true"`.  That seems a bit more explicit for error messages and then we can use `aria-describedby` for helper text only!

docs/components/field/demo/base-demo.md
NullVoxPopuli: I always go with PascalCase for components 🤷

ember-toucan-core/src/components/form/label.hbs
ynotdraw: Alright, I gave `aria-errormessage` a shot but ran into something similar to https://github.com/nvaccess/nvda/issues/8318 with voice over as well. The screenreader wouldn't announce the error message, but had no problems with the hint text.  I added two examples of wiring this up in the docs - one using `aria-describedby`  and the other using `aria-errormessage`.  If we can get `aria-errormesssage` to announce properly, I say we roll with that in our Form fields!

ember-toucan-core/package.json
```diff
@@ -57,9 +57,9 @@
     "@embroider/addon-dev": "^3.0.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
-    "@glint/core": "^0.9.7",
-    "@glint/environment-ember-loose": "^0.9.7",
-    "@glint/template": "^0.9.7",
+    "@glint/core": "^1.0.0-beta.3",
```
danwenzel, line 60: beta versions?

ember-toucan-core/src/components/form/error.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="type-xs-tight text-critical flex items-center mt-1" ...attributes>
+  {{! Icon taken from Tony's open source icon set, this is temporary! }}
```
danwenzel, line 2: temporary as in this will be updated in this PR?

danwenzel (approved):

ember-toucan-core/package.json
ynotdraw: Yeah! Pairing with @NullVoxPopuli we hit a few issues with the current version that this version seemed to solve.  By the time we release, maybe `@glint/core` will be proper 1.0.0?

NullVoxPopuli: yeah, the beta.3 is _very_ close to full 1.0.0

ember-toucan-core/src/components/form/error.hbs
ynotdraw: Excellent question!  Still _very_ temporary/WIP, but not worth blocking the PR on so we can continue to drive ahead from a development point of view.  We should have a much better understanding from a design perspective of _exactly_ what we want in about 2-4 weeks if I had to guess, but for the time being this is essentially a placeholder.  

We know we want an icon of sorts here, but not quite sure what it should be yet or exactly how it should look.  We think something like this works, but we definitely want to validate.  And it could be removed completely if we find it doesn't work!  Around the time mentioned above as well, we'll probably have quite a few PRs going up around styling tweaks to match what our design system expects; however, since we're re-thinking Form elements completely, things are still up in the air a bit.  We should land safely soon™ though 😄 .  We're definitely taking advantage of being pre-release 😂 !

ember-toucan-core/src/components/form/label.hbs
simonihmig: Hm, that's bad. Thanks for testing this, I should do the same for headless btw...

Probably have to change this to `describedby` as well in headless-form, tracking here: https://github.com/CrowdStrike/ember-headless-form/issues/47

Reminds me of the old days of browser incompatibilities, where you have nice new features in the spec, but you cannot use them due to IE support, instead going for crazy workarounds... 😩

ember-toucan-core/package.json
```diff
@@ -113,7 +113,12 @@
     "type": "addon",
     "main": "addon-main.cjs",
     "app-js": {
-      "./components/button.js": "./dist/_app_/components/button.js"
+      "./components/button.js": "./dist/_app_/components/button.js",
+      "./components/form/control.js": "./dist/_app_/components/form/control.js",
+      "./components/form/error.js": "./dist/_app_/components/form/error.js",
+      "./components/form/field.js": "./dist/_app_/components/form/field.js",
+      "./components/form/hint.js": "./dist/_app_/components/form/hint.js",
+      "./components/form/label.js": "./dist/_app_/components/form/label.js"
```
simonihmig, line 121: This is now making also the yielded components available for direct usage, as in `<Form::Label>`. Do we want this? We could also try to hide them as much as possible from users, and only make them available yielded by `<Form::Field>`. This is what headless is doing.

To do this, we would need to make tighten the glob(s) for appReexports  in the rollup config.

simonihmig (approved):
Approving, with just one minor comment to consider.

ember-toucan-core/package.json
ynotdraw: Perfect, thank you for the suggestion!  Went ahead and updated via https://github.com/CrowdStrike/ember-toucan-core/pull/27/commits/38e906ea0e5c44ceb68c3a22f6cd2ca27ef1b447 - let me know if I misconfigured something! 🙇   I verified I get an error on build when trying to import the Label component from the `test-app` with these changes.

I also stole your `'**/*.js',` public entry point line from `headless-form` as I think it makes sense for us too?

simonihmig: Perfect! :tada:

docs/components/field/demo/base-demo.md
nicolechung: But for the `yielded` part...here it is lowercase like `field`:

```
<Form::Field as |field|>
```

ynotdraw (approved):

PR #34, opened 2023-02-16
fix: Update glint registry with Form::Field

ynotdraw:
## 🐛  Description
Adds `Form::Field` to the Glint registry so we can use it internally.

Without this, we can't create components using the `Field` component or else we get a Glint error!

```bash

src/components/form/test.hbs:1:1 - error TS7053: Unknown name 'Form::Field'. If this isn't a typo, you may be missing a registry entry for this value; see the Template Registry page in the Glint documentation for more details.
  Element implicitly has an 'any' type because expression of type '"Form::Field"' can't be used to index type 'Globals'.
    Property 'Form::Field' does not exist on type 'Globals'.

  1 <Form::Field as |field|>
    ~~~~~~~~~~~~~~~~~~~~~~~~
  2   <field.Label for={{field.id}}>{{@label}}</field.Label>
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 21   {{/if}}
    ~~~~~~~~~
```

---

## 📚 Docs

- https://typed-ember.gitbook.io/glint/using-glint/ember/template-registry

---

## 🔬 How to Test

CI (Actions) should all be green 🟢 

---

N/A

### DIFF
--- a/ember-toucan-core/src/template-registry.ts
+++ b/ember-toucan-core/src/template-registry.ts
@@ -1,5 +1,7 @@
 import type ButtonComponent from './components/button';
+import type FieldComponent from './components/form/field';
 
 export default interface Registry {
   Button: typeof ButtonComponent;
+  'Form::Field': typeof FieldComponent;
 }
--- a/ember-toucan-core/unpublished-development-types/index.d.ts
+++ b/ember-toucan-core/unpublished-development-types/index.d.ts
@@ -3,8 +3,11 @@
 
 import '@glint/environment-ember-loose';
 
+import type FieldComponent from '../src/components/form/field';
+
 declare module '@glint/environment-ember-loose/registry' {
   export default interface Registry /* extends EmberPageTitle, etc, other addon registries */ {
     // local entries
+    'Form::Field': typeof FieldComponent;
   }
 }


ember-toucan-core/src/template-registry.ts
```diff
@@ -1,5 +1,7 @@
 import type ButtonComponent from './components/button';
+import type FieldComponent from './components/form/field';
 
 export default interface Registry {
   Button: typeof ButtonComponent;
+  'Form::Field': typeof FieldComponent;
```
ynotdraw, line 6: Admittedly still trying to wrap my head around the registry a bit, but I assume this template registry is for consumers of the add-on, like the `docs-app` and other apps+packages that'd consume this.

Adding this here, when I'd create a component inside our add-on using `Form::Field` I'd still get the glint error:

```bash

src/components/form/test.hbs:1:1 - error TS7053: Unknown name 'Form::Field'. If this isn't a typo, you may be missing a registry entry for this value; see the Template Registry page in the Glint documentation for more details.
  Element implicitly has an 'any' type because expression of type '"Form::Field"' can't be used to index type 'Globals'.
    Property 'Form::Field' does not exist on type 'Globals'.

  1 <Form::Field as |field|>
    ~~~~~~~~~~~~~~~~~~~~~~~~
  2   <field.Label for={{field.id}}>{{@label}}</field.Label>
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 21   {{/if}}
    ~~~~~~~~~
```

NullVoxPopuli (approved):

ember-toucan-core/unpublished-development-types/index.d.ts
```diff
@@ -3,8 +3,11 @@
 
 import '@glint/environment-ember-loose';
 
+import type FieldComponent from '../src/components/form/field';
+
 declare module '@glint/environment-ember-loose/registry' {
   export default interface Registry /* extends EmberPageTitle, etc, other addon registries */ {
     // local entries
+    'Form::Field': typeof FieldComponent;
```
ynotdraw, line 11: I'm not quite sure _why_ I need to add this here, I just found by adding it here as well the glint error (above) was resolved and things were happy.  I can only _assume_ we need this as the `template-registry` above and the registry here are "disconnected" as-in they don't know about each other.

nicolechung (approved):

ember-toucan-core/src/template-registry.ts
NullVoxPopuli: correct, this is what folks would import to add to their registry, like you would with unpublished-development-types (you could / should consume this template-registry in there)

ember-toucan-core/unpublished-development-types/index.d.ts
NullVoxPopuli: yeah -- the declare module is what glint actually uses for lookup -- you don't want to ship a declare module, because that breaks TS consumers

nicolechung: Do I need to add to both ^^ files as well when I add new components then?

simonihmig: Yes, exactly!

The key here is `declare module '@glint/environment-ember-loose/registry' { ... }`. This is what makes things known for the (classic non-strict-mode) template. And it _must_ exist only in one place for any app. So the app has the declaration, and there it can add its own type definitions and/or add existing registries imported from glint-enabled addons . See https://typed-ember.gitbook.io/glint/using-glint/ember/using-addons#using-glint-enabled-addons

So this file is indeed only local to the addon. The `template-registry.ts` doesn't do _anything_ (has no side-effect), as it doesn't have that `declare module '@glint/environment-ember-loose/registry'`. So only when this is imported from the consuming app and added to _its_ `@glint/environment-ember-loose/registry` will this have any effect.

ynotdraw: > yeah -- the declare module is what glint actually uses for lookup -- you don't want to ship a declare module, because that breaks TS consumers

Perfect, thanks so much for the clarification! Makes sense!

simonihmig (approved):

ember-toucan-core/src/template-registry.ts
ynotdraw: Perfect, thank ya

ember-toucan-core/unpublished-development-types/index.d.ts
ynotdraw: y'all are awesome, ty!

ynotdraw: > Do I need to add to both ^^ files as well when I add new components then?

@nicolechung I think it depends (based on my understanding of above):

- If your component is **only** going to be used externally, putting it in `template-registry` is just fine
- If your component is going to be used _internally_ to toucan-core, then you'll need it in `unpublished-development-types/index.d.ts` (say, as a building-block component for other components in toucan-core)
- If it's going to be used _both_ internally and externally then both places

nicolechung: Thank you for the clarification!

PR #35, opened 2023-02-16
[WIP] 30 create input field

nicolechung:
Currently: a very early draft version.

Todos
- [ ] clean up git history with proper commit messages
- [ ] Add images of docs
- [ ] docs: add error example
- [ ] docs: add focus state example
- [ ] docs: add `number` type example
- [ ] docs: add `time` example
- [ ] docs: add `week` example
- [ ] docs: add `month` example
- [ ] tests: it renders with label and hint
- [ ] tests: it renders with proper error message
- [ ] tests: it can have `...attributes`


ember-toucan-core/src/components/form/input-field.hbs
```diff
@@ -0,0 +1,26 @@
+<div class="flex flex-col {{if @isDisabled 'text-disabled cursor-not-allowed'}}">
```
nicolechung, line 1: Note: none of this is the finished styling currently, as we are waiting on user and design research. Styling will be for an upcoming iteration.

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,90 @@
...
+      <InputField
+        @label="Label"
+        type="text"
+        @error="There is an error"
+        />
+    </template>);
+
+    const label = 'label'
+    const input = 'input';
+    const error = 'div.text-critical';
```
nicolechung, line 39: **Question:** wondering if I should add `data-test` selectors in the https://github.com/CrowdStrike/ember-toucan-core/pull/35/files#diff-a8dfe5662c125f37bbd67d22d9112a123c0ffac52f7254a0fa217dd019452071 to get around these horribly specific class selectors?

ember-toucan-core/src/components/form/input-field.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <Form::Field as |field|>
+    <field.Label for={{field.id}}>{{@label}}</field.Label>
+
+    {{#if @hint}}
+      <field.Hint id={{field.hintId}}>{{@hint}}</field.Hint>
+    {{/if}}
+
+    <field.Control class="flex">
+      <input
+        class="m-1 h-6 w-0 flex-grow bg-transparent p-1
```
ynotdraw, line 13: Do we want the margin here via `m-1`? That'll push elements around a bit.

I tried getting _pretty_ close to what we are going for in with the textarea, which I think is what we'll mostly want here https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-0cf8ecb43c00a4f64c9e8710b853499367e0130793dffdb772c0bb4ef9a4dd0eR13


Something like this may be close:
```hbs
class="block h-6 p-1 bg-overlay-1 focus:outline-none rounded-sm shadow-focusable-outline focus:shadow-focus-outline {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
```

I also left off other soon™ disabled styles, but happy to align with whatever so we remain consistent as to make it easier on us later when we go and update the styles to match post-testing!

ember-toucan-core/src/components/form/input-field.ts
```diff
@@ -0,0 +1,18 @@
+import Component from '@glimmer/component';
+
+interface ToucanFormInputFieldComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    label?: string;
```
ynotdraw, line 6: Should we make `label` required for a11y reasons? (Even if we eventually allow it to be visually hidden)

```suggestion
    label: string;
```

ember-toucan-core/src/components/form/input-field.ts
```diff
@@ -0,0 +1,18 @@
...
+    isDisabled?: boolean;
+    readonly?: boolean;
+  };
+  Blocks: {
+    default: [];
+  };
+}
+
+// eslint-disable-next-line ember/no-empty-glimmer-component-classes
+export default class ToucanFormInputFieldComponent extends Component<ToucanFormInputFieldComponentSignature> {}
```
ynotdraw, line 18: If you're not using a constructor here, you could do this instead:

```js
import templateOnlyComponent from '@ember/component/template-only';

export default templateOnlyComponent<ToucanFormInputFieldComponentSignature>();
```

but if we do plan on adding a constructor eventually it's probably moot

pnpm-lock.yaml
```diff
@@ -8607,7 +8607,7 @@ packages:
       debug: 4.3.4
       enhanced-resolve: 5.12.0
       eslint: 8.33.0
-      eslint-plugin-import: 2.27.5_fhrhggiruboehhi74bczdxbcti
+      eslint-plugin-import: 2.27.5_nowqz4jutkd4a233czbfk7jsgu
```
ynotdraw, line 8610: Do you know why this changed by chance? Since dependencies weren't adjusted in a `package.json` I wouldn't have expected any `ppm-lock.yaml` file changes

docs/components/input-field/index.md
```diff
@@ -0,0 +1,133 @@
...
+  @label="Label"
+  @hint="extra information about the field"
+  type="text"
+  @isDisabled={{true}}
+/>
+```
+
+## UI States
+
+### InputField with Label
```
ynotdraw, line 94: I like the named headings here for UI states.  I'll update my PR to follow!

(For others: @nicolechung and I brainstormed how this may be a good idea if we want to add visual regression testing.  Then we can see the component in every expected state and visually diff things)

ynotdraw (commented):
All pretty minor comments -  This is exciting to see come together 🎉  🚀 .  Happy to pair tomorrow as well if we want to land on some consistency ideas/templates between our two PRs!

ember-toucan-core/src/components/form/input-field.hbs
```diff
@@ -0,0 +1,32 @@
...
+            @isDisabled
+            'text-disabled cursor-not-allowed'
+            'text-titles-and-attributes'
+          }}
+          "
+        id={{field.id}}
+        aria-describedby="{{if @hint field.hintId}} {{if @error field.errorId}}"
+        aria-invalid={{if @error "true"}}
+        disabled={{@isDisabled}}
+        ...attributes
```
ynotdraw, line 24: Kind of wondering if it's worth it for us to set `type="text"` by default?  Then folks can override since we `...attributes`.  What do you think?

simonihmig (commented):
I saw the request for review here only after the request for #36, so reviewed that first. But a couple of comments I left there apply to this in similar ways. So I am not duplicating them here. Instead maybe we could discuss them over there, and depending on the outcomes apply changes (if there are any) here accordingly?

ember-toucan-core/src/components/form/input-field.ts
nicolechung: I totally missed this from the notes...need to add an assert here too (and test for it)

ember-toucan-core/src/components/form/input-field.ts
nicolechung: This should be a class now

pnpm-lock.yaml
nicolechung: Maybe it's because I did `pnpm i` without `--frozen-lockfile`??

https://pnpm.io/cli/install#--frozen-lockfile

nicolechung: I thought that was only needed in CI though

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,138 @@
...
+    assert.dom(hint).exists('Expected to have hint component rendered');
+    assert.dom(hint).hasText('Hint text visible here', 'Expected to have hint text "error"');
+
+    if (id) {
+     assert.dom(hint).hasAttribute('id', id);
+    }
+  });
+
+  test('it accepts @value and @onChange', async function (assert) {
+    const onChangeCallback = sinon.spy();
```
ynotdraw, line 108: Instead of adding `sinon` as a dependency you can use `assert.step` instead! https://github.com/CrowdStrike/ember-toucan-core/blob/main/test-app/tests/integration/components/button-test.gts#L131

ember-toucan-core/src/components/form/controls/input.ts
```diff
@@ -0,0 +1,24 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+import { action } from '@ember/object';
+
+interface ToucanFormControlsInputComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    isDisabled?: boolean;
```
nicolechung, line 8: If anyone has any advice as to why this is erroring: https://github.com/CrowdStrike/ember-toucan-core/actions/runs/4247201741/jobs/7384984044 would appreciate it!

docs/components/input-field/index.md
```diff
@@ -0,0 +1,133 @@
+# Input field
```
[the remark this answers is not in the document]
nicolechung: I can adjust my "field" doc to match yours...let me take a look.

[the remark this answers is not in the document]
nicolechung: Should be updated now in this commit: https://github.com/CrowdStrike/ember-toucan-core/pull/35#:~:text=match%20TextareaField%20docs-,6428f0f,-fix%3A%20Input%20Control

nicolechung:
> review here only after the request for #36, so reviewed that first. But a couple of comments I left there apply to this in similar ways. So I am not duplicating them here. Instead maybe we could discuss them over there, and depending on the outcomes apply changes (if there are any) here accordingly?

I ended up pairing with @ynotdraw so hopefully this addresses the issues mentioned.

test-app/package.json
```diff
@@ -112,6 +112,7 @@
     "prettier-plugin-ember-template-tag": "^0.3.0",
     "qunit": "^2.19.3",
     "qunit-dom": "^2.0.0",
+    "sinon": "^15.0.1",
```
ynotdraw, line 115: We shouldn't need this anymore which should remove it from the lock file as well! 🎉

docs/components/input/index.md
```diff
@@ -0,0 +1,40 @@
+# Input 
+
+Provides a Toucan-styled [textarea element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea). If you are building forms, you may be interested in the TextareaField component instead.
```
ynotdraw, line 3:
```suggestion
Provides a Toucan-styled [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input). If you are building forms, you may be interested in the InputField component instead.
```

docs/components/input/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Value
+
+To set the `value` attribute of the `<textarea>`, provide `@value`.
```
ynotdraw, line 7:
```suggestion
To set the `value` attribute of the `<input>`, provide `@value`.
```

ember-toucan-core/src/components/form/input-field.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @action
+  handleInput(e: Event | InputEvent): void {
+    assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
+
+    if (this.args.onChange) {
+      this.args.onChange(e.target.value, e);
+    }
+  }
```
ynotdraw, line 38: Do we still need this here since we can pass the argument directly through to the underlying Input component?

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+
+    assert.verifySteps([]);
+
+    const input: HTMLInputElement | null  = document.querySelector('[data-input]');
+
+    if (!input) {
+      throw new Error('an input was not detected');
+    }
+    
+    assert.strictEqual(input.value, 'Avocado', 'input has the set @value');
```
simonihmig, line 139: I would suggest to use `qunit-dom` here, which would also remove need to manually query the element. See https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-b7acb71cb9e9389fb6732d2b7eb21e52658721f8307ab8c972d27de292a9adb3R130

test-app/tests/integration/components/input-test.gts
```diff
@@ -0,0 +1,84 @@
...
+    assert
+      .dom('[data-input]')
+      .doesNotHaveClass('text-titles-and-attributes');
+  });
+
+  test('it spreads attributes to the underlying textarea', async function (assert) {
+    await render(<template>
+      {{! we do not require a label, but instead suggest using Field / TextareaField }}
+      {{! template-lint-disable require-input-label }}
+      <InputControl placeholder="Placeholder text" data-input />
```
simonihmig, line 44: Interesting, how does template-lint know this is representing an `<input>`? By the component name? The component could render something totally different, where that rule does not apply... 🤔

simonihmig (commented):
Looks very good overall! Just some minor things to address...

ember-toucan-core/rollup.config.mjs
```diff
@@ -24,7 +24,11 @@ export default {
     // These are the modules that should get reexported into the traditional
     // "app" tree. Things in here should also be in publicEntrypoints above, but
     // not everything in publicEntrypoints necessarily needs to go here.
-    addon.appReexports(['components/*.js', 'components/form/*.js']),
+    addon.appReexports([
+      'components/*.js',
+      'components/form/*.js',
+      'components/form/controls/*.js',
+    ]),
```
ynotdraw, line 31: If you rebase, you won't need to adjust the rollup config any longer!

(For ref: https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-ef1c3fa189e4cf0ff2a244e5f3ad25086f8d12247f8a37c66e2dc1a8aa8cebb7)

ember-toucan-core/src/components/form/controls/input.hbs
```diff
@@ -0,0 +1,8 @@
+<input
+  class="bg-overlay-1 focus:outline-none shadow-focusable-outline focus:shadow-focus-outline block rounded-sm p-1
+    {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
```
ynotdraw, line 3: I found out yesterday we need a `@hasError` argument so we can conditionally apply the error shadow.  Something  like this should work:

```suggestion
  class="bg-overlay-1 focus:outline-none focus:shadow-focus-outline block rounded-sm p-1
    {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}
    {{if @hasError 'shadow-error-outline' 'shadow-focusable-outline'}}"
```

(For reference:  https://github.com/CrowdStrike/ember-toucan-core/pull/52/files#diff-f86c0f8f074d8a714e4e434f71ef3d3ee3db10702a7260cce69f4143ab2b71f7)

ember-toucan-core/src/components/form/controls/input.ts
```diff
@@ -0,0 +1,24 @@
...
+    onChange?: (value: string, e: Event | InputEvent) => void;
```
ynotdraw, line 9: I added a re-usable type here if you want to give it a go! 

```suggestion
    onChange?: OnChangeCallback<string>;
```

(Reference: https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-90b0ef8397a4c8dabf3349619492464098ba317ce1542a255dcc35ac8640615f)

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+    </template>);
+
+    const label = '[data-label]'
+    const input = '[data-input]';
+
+    assert.dom(label).exists('Expected to have label block rendered');
+    assert.dom(label).hasText('Label', 'Expected to have label text "label"');
+
+    assert.dom(input).exists('Expected to have input tag rendered');
+    assert.dom(input).hasAttribute('type', 'text');
```
ynotdraw, line 29: We may want to verify the `id` gets set here as well:

```js
assert.dom('[data-input]').hasAttribute('id');
```

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+        @error="There is an error"
+        data-input
+        />
+    </template>);
+
+    const label = '[data-label]'
+    const input = '[data-input]';
+    const error = '[data-error]';
+   
+    const inputElement = document.querySelector('[data-input]');
```
ynotdraw, line 61: Instead of `document.querySelector` you could use `find`

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+    assert.dom(label).hasText('Label', 'Expected to have label text "label"');
+
+    assert.dom(input).exists('Expected to have input tag rendered');
+    assert.dom(input).hasAttribute('type', 'text');
+   
+    assert.dom(error).exists('Expected to have error component rendered');
+    assert.dom(error).hasText('There is an error', 'Expected to have error text "error"');
+
+    if (id) {
+      assert.dom(error).hasAttribute('id', id);
```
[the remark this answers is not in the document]
ynotdraw: yup yup.  One thing I did for Textarea to assert this stuff: https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-b7acb71cb9e9389fb6732d2b7eb21e52658721f8307ab8c972d27de292a9adb3R63

```js
     let error = find('[data-error]');

     assert.dom(error).hasText('Error text');
     assert.dom(error).hasAttribute('id');

     let errorId = error?.getAttribute('id') || '';
     assert.ok(errorId, 'Expected errorId to be truthy');

     let describedby =
       find('[data-textarea]')?.getAttribute('aria-describedby') || '';
     assert.ok(
       describedby.includes(errorId),
       'Expected errorId to be included in the aria-describedby'
     );

     assert.dom('[data-textarea]').hasAttribute('aria-invalid', 'true');
```

Then no conditionals and defaulting to `|| ''` and asserting it's truthy

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+    assert.dom(label).hasText('Label', 'Expected to have label text "label"');
+
+    assert.dom(input).exists('Expected to have input tag rendered');
+    assert.dom(input).hasAttribute('type', 'text');
+   
+    assert.dom(hint).exists('Expected to have hint component rendered');
+    assert.dom(hint).hasText('Hint text visible here', 'Expected to have hint text "error"');
+
+    if (id) {
+     assert.dom(hint).hasAttribute('id', id);
```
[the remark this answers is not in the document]
ynotdraw: For reference here's how I handled this in Textarea: https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-b7acb71cb9e9389fb6732d2b7eb21e52658721f8307ab8c972d27de292a9adb3R42

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+
+    assert.dom(input).exists('Expected to have input tag rendered');
+    assert.dom(input).hasAttribute('type', 'text');
+   
+    assert.dom(error).exists('Expected to have error component rendered');
+    assert.dom(error).hasText('There is an error', 'Expected to have error text "error"');
+
+    if (id) {
+      assert.dom(error).hasAttribute('id', id);
+    }
```
ynotdraw, line 76: May be good to also assert `aria-invalid` is set on the `input`

```suggestion
    }
    assert.dom('[data-input]').hasAttribute('aria-invalid', 'true');
```

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,148 @@
...
+    
+    assert.strictEqual(input.value, 'Avocado', 'input has the set @value');
+
+    await fillIn('[data-input]', 'Banana');
+
+    assert.strictEqual(input.value, 'Banana', 'input has the set @value');
+
+    assert.verifySteps(['handleChange']);
+  });
+
```
ynotdraw, line 147: Not sure if overkill, but I added a test for making sure _both_ hint and error IDs are set properly on `aria-describedby`

(Reference: https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#diff-b7acb71cb9e9389fb6732d2b7eb21e52658721f8307ab8c972d27de292a9adb3R81)

test-app/tests/integration/components/input-test.gts
```diff
@@ -0,0 +1,84 @@
...
+      {{! template-lint-disable require-input-label }}
+      <InputControl data-input />
+    </template>);
+
+    assert.dom('[data-input]').hasTagName('input');
+    assert.dom('[data-input]').hasClass('text-titles-and-attributes');
+    assert.dom('[data-input]').doesNotHaveClass('text-disabled');
+  });
+
+  test('it disables the textarea using `@isDisabled`', async function (assert) {
```
ynotdraw, line 26:
```suggestion
  test('it disables the input using `@isDisabled`', async function (assert) {
```

test-app/tests/integration/components/input-test.gts
```diff
@@ -0,0 +1,84 @@
...
+    </template>);
+
+    assert.dom('[data-input]').isDisabled();
+    assert.dom('[data-input]').hasClass('text-disabled');
+    assert
+      .dom('[data-input]')
+      .doesNotHaveClass('text-titles-and-attributes');
+  });
+
+  test('it spreads attributes to the underlying textarea', async function (assert) {
```
ynotdraw, line 40:
```suggestion
  test('it spreads attributes to the underlying input', async function (assert) {
```

test-app/tests/integration/components/input-test.gts
ynotdraw: Oh interesting! I had to do the same thing with Textarea - may you're right that it is by the component name. I'd be curious!  I'd also be impressed if it dug into the template to find it was in fact an `<input>`!

test-app/tests/integration/components/input-test.gts
```diff
@@ -0,0 +1,84 @@
...
+    </template>);
+
+    assert
+      .dom('[data-input]')
+      .hasAttribute('placeholder', 'Placeholder text');
+  });
+
+  test('it sets the value attribute via `@value`', async function (assert) {
+    await render(<template>
+      {{! we do not require a label, but instead suggest using Field / TextareaField }}
```
ynotdraw, line 54:
```suggestion
      {{! we do not require a label, but instead suggest using Field / InputField }}
```

ynotdraw (commented):
Apologies for more comments! Changed a few things on the Textarea PR before merging!

test-app/tests/integration/components/input-test.gts
nicolechung: >Interesting, how does template-lint know this is representing an <input>? By the component name? The component could render something totally different, where that rule does not apply... 🤔

I think it visits all ElementNodes and looks for any instance of say, `Input` or `input`...if it sees `input` it then checks the parent's attributes.
Screen Shot 2023-02-27 at 1 25 19 PM


https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/require-input-label.js#L88

ember-toucan-core/src/components/form/controls/input.hbs
nicolechung: Interesting...are we always going to conditionally apply via a `@hasError` argument or should it be `this.hasError` to account for if there are arguments passed to `@error`?

nicolechung: oh nvm `@error` comes from the InputField component

ember-toucan-core/src/components/form/input-field.ts
nicolechung: @ynotdraw remove the `handleInput` (or the entire backing class) you mean?

test-app/app/templates/application.hbs
```diff
@@ -1,3 +1,45 @@
 <h2 id="title">Welcome to Ember</h2>
 
-{{outlet}}
\ No newline at end of file
+{{!-- template-lint-disable require-input-label --}}
```
nicolechung, line 3: Screen Shot 2023-02-27 at 2 36 30 PM

Just note: I added this, but did not install the related styles. If we need that maybe it can be a follow up?

test-app/tests/integration/components/input-field-test.gts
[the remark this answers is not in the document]
nicolechung: Question: if both the `errorId` and `describedby` end up being `''`, is that assertion correct? Or that case won't happen because assert.ok tests for truthiness? (is that the "errorId" is truthy and `describedby` is truthy??)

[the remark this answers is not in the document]
nicolechung: I'm still going to use that snippet btw.

test-app/tests/integration/components/input-test.gts
simonihmig: Yeah, but our tag here is `InputControl`, which is not part of [this SET](https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/rules/require-input-label.js#L29). So still seems weird to me. 🤔 

Anyway, nothing that should block us here, just curious...

docs/components/input/index.md
```diff
@@ -0,0 +1,40 @@
+# Input 
+
+Provides a Toucan-styled [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input). If you are building forms, you may be interested in the TextareaField component instead.
```
simonihmig, line 3:
```suggestion
Provides a Toucan-styled [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input). If you are building forms, you may be interested in the InputField component instead.
```

ember-toucan-core/unpublished-development-types/index.d.ts
```diff
@@ -11,5 +12,6 @@ declare module '@glint/environment-ember-loose/registry' {
     // local entries
     'Form::Field': typeof FieldComponent;
     'Form::Controls::Textarea': typeof TextareaControlComponent;
+    'Form::Controls::Input': typeof InputControlComponent;
```
simonihmig, line 15: Btw, in our _internal_ types we can also consume our own public registry, so we don't have to add this twice every time. See https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/unpublished-development-types/index.d.ts#L32

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,165 @@
...
+      <InputField
+        @label="Label"
+        type="text"
+        data-input
+        />
+    </template>);
+
+    const label = '[data-label]'
+    const input = '[data-input]';
+    const inputId = find(input)?.getAttribute('id') || '';
```
simonihmig, line 24: When the input has an ID of an empty string (and having `<label for="">`), then I think this would make this test pass. Better to assert also for the existence of a proper ID:

```suggestion
    const inputId = find(input)?.getAttribute('id');
    
    assert.ok(inputId, 'Expected to have ID');
```

test-app/tests/integration/components/input-test.gts
nicolechung: I'm not totally understanding what's going on in the lint rule code, but [in the tests](https://github.com/ember-template-lint/ember-template-lint/blob/master/test/unit/rules/require-input-label-test.js#L74) it looks like it searches for any `input` and checks the surrounding (parent?) tags to see if there is an `label`. With `InputControl` it can tell there is an `input` inside but no label.

test-app/tests/integration/components/input-field-test.gts
nicolechung: okay I think this is here, but you might want to confirm since there is no "Outdated" label

nicolechung: Okay yeah it's here: https://github.com/CrowdStrike/ember-toucan-core/pull/35/files#diff-b1d74a6d42d39cd9e5f8eb62146064c2694715629721f598278442b2a9249855R31

test-app/tests/integration/components/input-field-test.gts
nicolechung: I added it here: https://github.com/CrowdStrike/ember-toucan-core/pull/35/files#diff-b1d74a6d42d39cd9e5f8eb62146064c2694715629721f598278442b2a9249855R108

test-app/tests/integration/components/input-field-test.gts
nicolechung: Screen Shot 2023-02-28 at 12 18 34 PM

The linter was complaining about the line change on 24 but I added the other suggestion - thank you!

pnpm-lock.yaml
nicolechung: any ideas @ynotdraw ? I wonder (since I removed the dependency) if I should just revert this change. Let me try that...

ember-toucan-core/src/components/form/input-field.ts
nicolechung: oh wait I see...going to remove the backing class.

ember-toucan-core/unpublished-development-types/index.d.ts
nicolechung: I think @ynotdraw PR is going to get merged first, will pull that from there: https://github.com/CrowdStrike/ember-toucan-core/pull/50

docs/components/input/index.md
```diff
@@ -0,0 +1,40 @@
...
+  handleChange(value, e) {
+    console.log({ e, value });
+    this.value = value;
+  }
+}
+```
+
+## Disabled State
+
+Set the `@isDisabled` argument to disable the `<input>`.
```
ynotdraw, line 40: Nit: May want to add info about `@hasError` :


```suggestion
Set the `@isDisabled` argument to disable the `<input>`.

## Error State

Set the `@hasError` argument to apply an error box shadow to the `<input>`.
```

ember-toucan-core/src/components/form/controls/input.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import type { OnChangeCallback } from '../../../-private/types';
+
+interface ToucanFormControlsInputComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    hasError?: boolean;
+    isDisabled?: boolean;
+    onChange?: OnChangeCallback<string>;
+    readonly?: boolean;
```
ynotdraw, line 13: Nit: it looks like we aren't using this anywhere yet

ember-toucan-core/src/components/form/input-field.hbs
```diff
@@ -0,0 +1,27 @@
+<div
+  class="flex flex-col {{if @isDisabled 'text-disabled cursor-not-allowed'}}"
```
ynotdraw, line 2: Is the `cursor-not-allowed` class something we want here?

ember-toucan-core/src/components/form/input-field.ts
```diff
@@ -0,0 +1,29 @@
...
+import { assert } from '@ember/debug';
+
+interface ToucanFormInputFieldComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    error?: string;
+    label: string;
+    hint?: string;
+    isDisabled?: boolean;
+    onChange?: (value: string, e: Event | InputEvent) => void;
```
ynotdraw, line 11: Nit: To keep this in sync with the underlying component API you could do:

```suggestion
    // import ToucanFormInputControlComponentSignature
    onChange?: ToucanFormInputControlComponentSignature['Args']['onChange'];
```

or

```suggestion
    // import OnChangeCallback
    onChange?: OnChangeCallback<string>;
```

ember-toucan-core/src/components/form/input-field.ts
```diff
@@ -0,0 +1,29 @@
...
+    readonly?: boolean;
```
ynotdraw, line 12: Same question here about `readonly` - do we want to support this yet?

ember-toucan-core/unpublished-development-types/index.d.ts
ynotdraw: Yup! Should be able to rebase and then not have to modify this file! 🎉

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,166 @@
...
+        data-input
+        />
+    </template>);
+
+    const label = '[data-label]'
+    const input = '[data-input]';
+    const error = '[data-error]';
+  
+    assert.dom(label).exists('Expected to have label component rendered');
+    assert.dom(label).hasText('Label', 'Expected to have label text "Label"');
```
ynotdraw, line 65: Nit: Up to you, but there's a test above validating `label`, so you could probably get away with this test focusing on only testing that the error displays and removing these assertions as they're redundant (same for other tests).

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,166 @@
...
+
+    assert.dom(input).exists('Expected to have input tag rendered');
```
ynotdraw, line 67: Similar note here, but totally up to you.  The `it renders` test above asserts that the input is there, so keeping the test focused on error / hint states would make the tests smaller and a bit easier to follow potentially.

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,166 @@
...
+      throw new Error('an input was not detected');
+    }
+   
+    assert.dom(input).hasValue('Avocado', 'input has the original value');
+
+    await fillIn('[data-input]', 'Banana');
+
+    assert.dom(input).hasValue('Banana', 'input has the set @value');
+
+    assert.verifySteps(['handleChange']);
```
ynotdraw, line 163: Nit / no need to change: this could maybe be simplified with:

```suggestion
    assert.dom('[data-input]').hasValue('Avocado', 'input has the original value');

    await fillIn('[data-input]', 'Banana');

    assert.dom('[data-input]').hasValue('Banana', 'input has the set @value');

    assert.verifySteps(['handleChange']);
```

If `fillIn` doesn't find an element or if the value isn't set as expected you'd get an assertion error

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -0,0 +1,166 @@
...
+  });
+
+});
```
ynotdraw, line 166: Nit: May be good as well to add a test for the `rootTestSelector` to make sure we have tests for all of our args:

```suggestion
    test('it applies the provided `@rootTestSelector` to the data-root-field attribute', async function (assert) {
    await render(<template>
      <InputField
        @label="Label"
        @rootTestSelector="selector"
        data-input
      />
    </template>);

    assert.dom('[data-root-field="selector"]').exists();
  });
});
```

ember-toucan-core/src/components/form/input-field.hbs
nicolechung: Will remove it for now...I was thinking we wanted to whole input field to be disabled but probably better to adjust this (if even needed!) when design is more finalized.

BTW (not sure if this is clear or not) `cursor-not-allowed` changes the cursor style, it doesn't hide it. https://tailwindcss.com/docs/cursor

Unlike what the naming suggests...

ynotdraw (approved):
🚀 nice work!

simonihmig (approved):

Issue #31, opened 2023-02-13
Create TextAreaField Component


ynotdraw:
Depends on https://github.com/CrowdStrike/ember-toucan-core/issues/10

As a developer, I'd like to have a TextAreaField Component, so that I may have standardized styles without styling the component myself.  It should build on top of `<Field` introduced in the above issue.

**Acceptance criteria**
- TextArea has a label argument
- TextArea has a hint argument that can be displayed optionally.
- TextArea has a control argument that can accept different textarea attributes
- TextArea has an error argument
- Error text is only displayed during errors

PR #36, opened 2023-02-17
#31: Add Textarea component

ynotdraw:
## 🚀 Description
This PR introduces the `TextareaField` component.

**NOTE:** These components are not 100% styled yet, as we are still figuring out where to land on a few things.  This is about ~75% close though.

---

## 🔬 How to Test

- These are only presentational components, so test should pass
- View the link at https://90c89336.ember-toucan-core.pages.dev/docs/components/textarea-field
  - **NOTE:** These components are not 100% styled yet, as we are still figuring out where to land on a few things.  This is about ~75% close though.

---

Screenshot 2023-02-21 at 2 38 59 PM

Screenshot 2023-02-21 at 2 38 54 PM


docs/components/textarea-field/demo/base-demo.md
```diff
@@ -0,0 +1,32 @@
...
+  updateValue(e) {
+    this.value = e.target.value;
+
+    if (e.target.value !== 'textarea') {
+      this.errorMessage = 'Input must match "textarea"';
+      return;
+    }
+
+    this.errorMessage = null;
+  }
```
ynotdraw, line 30: Something like this _may_ be overkill for a demo?

docs/components/textarea-field/index.md
```diff
@@ -0,0 +1,48 @@
...
+Set the `@isDisabled` argument to disable the `<textarea>`.
+
+## Attributes and Modifiers
+
+Consumers have direct access to the underlying [textarea element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea), so all attributes are supported. Modifiers can also be added directly to the textarea as shown in the demo.
+
+## All UI States
+
+<div class="flex flex-col space-y-4">
+<Form::TextareaField
```
simonihmig, line 28: Uh, wat? You can just insert components into markdown with docfy? TIL 🤯

ember-toucan-core/src/components/form/textarea-field.hbs
```diff
@@ -0,0 +1,28 @@
...
+        id={{field.id}}
+        ...attributes
+      />
+    </field.Control>
+
+    {{#if @error}}
+      <field.Error
+        id={{field.errorId}}
+        data-textarea-error
+      >{{@error}}</field.Error>
```
simonihmig, line 25: This assumes `@error` is always a simple (escaped) string, no further HTML markup. Which is not wrong, if that is what we want.

headless-form does allow for more than one validation error though. In terms of UX we can decide to
* only show the first error (which is a valid approach IMO, and would work fine with the current implementation here)
* show all errors, rendered with `<br/>` or a full `<ul>` list maybe? In which case we wouldn't be able to support this with a simple `@error` I think?

To support the second case (if we want that), we could
* conditionally yield to a named block, like `{{#if (has-block "error")}}{{yield to="error"}}{{(/if}}`, letting the user decide how to render the list of errors
* or let `@error` also be an array, detect that here, and `{{#each}}` over them here, with some opinionated way to render them (`<br/>` vs. `<ul>` vs. whatever). Which is ok (maybe better), since we are representing a design system here, so can impose the way it should look

Thoughts?

ember-toucan-core/src/components/form/textarea-field.hbs
```diff
@@ -0,0 +1,28 @@
+<div class="flex flex-col">
+  <Form::Field as |field|>
+    <field.Label for={{field.id}} data-textarea-label>{{@label}}</field.Label>
+
+    {{#if @hint}}
+      <field.Hint id={{field.hintId}} data-textarea-hint>{{@hint}}</field.Hint>
+    {{/if}}
+
+    <field.Control>
+      <textarea
```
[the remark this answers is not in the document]
ynotdraw: I am glad you brought this up! I was thinking the exact same but then I thought maybe I'm overthinking / over solving and decided not to do it. 

I agree though, I think it makes sense so that folks can only use those individual atoms as needed and they're styled as Toucan components.

I'll get these changes added! 🙇

ember-toucan-core/src/components/form/textarea-field.hbs
```diff
@@ -0,0 +1,28 @@
+<div class="flex flex-col">
+  <Form::Field as |field|>
+    <field.Label for={{field.id}} data-textarea-label>{{@label}}</field.Label>
```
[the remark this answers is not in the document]
ynotdraw: Excellent points.  Yeah this could get tricky.  I wasn't quite sure either.  I think it's important for us to splat the attributes onto the underlying control so that folks can add modifiers, attributes, etc., but then that makes access child elements of the Field a lot trickier.

So are you thinking something like:

```
<div class="flex flex-col" data-test-selector={{@testSelector}}>
   <Form::Field as |field|>
     <field.Label for={{field.id}} data-label>{{@label}}</field.Label>
```

As you say, then we could allow for selectors like `[data-test-selector="Simon"] [data-label]`.  Maybe we should provide a fractal-page-object to help with some of this?

ember-toucan-core/src/components/form/textarea-field.hbs
```diff
@@ -0,0 +1,28 @@
...
+      <field.Hint id={{field.hintId}} data-textarea-hint>{{@hint}}</field.Hint>
+    {{/if}}
+
+    <field.Control>
+      <textarea
+        aria-invalid={{if @error "true"}}
+        aria-describedby="{{if @error field.errorId}} {{if @hint field.hintId}}"
+        class="block h-20 min-h-6 p-1 bg-overlay-1 focus:outline-none rounded-sm shadow-focusable-outline focus:shadow-focus-outline
+          {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
+        disabled={{@isDisabled}}
```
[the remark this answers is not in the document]
ynotdraw: For sure - those should totally work! I think for this case it _somewhat_ depends.  We are testing out some designs where the label and hint text may also look different based on the disabled state.  If we do in fact change the style of these surrounding elements as well, I had thought that maybe a component argument would be easiest to support as then we can check `@isDisabled` rather than having to reach into the `<textarea>` and pull out the disabled attribute; however, if we are **not** going that route and _only_ styling the disabled `<textarea>`, then I 100% agree with using the `disabled:` selector for sure.

So for example:

```hbs
<field.Label class="{{if @isDisabled 'some-custom-styling'}}" for={{field.id}} data-textarea-label>{{@label}}</field.Label>
```

So I guess, at this point in time, we aren't quite sure exactly where we will land from a styling point here, so I think we have some flexibility for the time being.  We have another task to circle back through all of the components and update them to match designs post-usability testing though.  We are probably 2-3 weeks away from this happening though.  What do you feel is best here knowing above?  I'm still on the fence!  Could go either way.

ember-toucan-core/src/components/form/textarea-field.hbs
```diff
@@ -0,0 +1,28 @@
...
+        id={{field.id}}
+        ...attributes
```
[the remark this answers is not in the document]
ynotdraw: Ah! Good call.  I was initially thinking the consumers would control all of that, but now I remember controlled vs uncontrolled components in React and it seems like we should definitely set `value` for input, textarea, etc. and have an internal `setValue` like your link and this https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/control/textarea.ts#L43 right?

ember-toucan-core/src/components/form/textarea-field.hbs
ynotdraw: > conditionally yield to a named block, like {{#if (has-block "error")}}{{yield to="error"}}{{(/if}}, letting the user decide how to render the list of errors

I'm glad you mentioned this! I was thinking even yielding to our other blocks as well, hint and label to allow a bit more flexibility with things.  What do you think?  I like this idea though, exposing a block for errors.

> or let @error also be an array, detect that here, and {{#each}} over them here, with some opinionated way to render them (<br/> vs. <ul> vs. whatever). Which is ok (maybe better), since we are representing a design system here, so can impose the way it should look

I personally quite like this idea!  I'm curious from a design perspective how multiple errors could look though.  Maybe I'll ask in our channel about it and see what folks think then go from there?

docs/components/textarea-field/index.md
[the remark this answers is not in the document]
ynotdraw: Haha yeah I kind of just stumbled upon this working 🙈 😂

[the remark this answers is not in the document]
nicolechung: I also just picked it up from @ynotdraw too...

Looks like the magic for this is coming from here: https://github.com/josemarluedke/remark-hbs

ember-toucan-core/src/components/form/textarea-field.hbs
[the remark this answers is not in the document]
nicolechung: I should do this for InputField also.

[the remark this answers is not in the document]
simonihmig: Yeah, right. `@value` is the one thing, but `@onChange` (or however you want to name it) is also important. First I thought users could just apply `{{on ...}}` themselves, but indeed the action you linked to shows an important point: it would be way more convenient to not have to deal with getting the (changed) value from `event.target`!

ember-toucan-core/src/components/form/textarea-field.hbs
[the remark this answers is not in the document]
simonihmig: yeah, this could work.

> Maybe we should provide a fractal-page-object to help with some of this?

Hm, maybe? I haven't yet touched fractal page objects since joining CS, so can't really speak to it. Are we providing these in other OSS addons, or are they very specific to CS?

ember-toucan-core/src/components/form/textarea-field.hbs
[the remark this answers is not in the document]
simonihmig: > If we do in fact change the style of these surrounding elements as well, I had thought that maybe a component argument would be easiest to support

Yeah, makes sense in this case!

> having to reach into the <textarea> and pull out the disabled attribute

No, we should _definitely_ not do that! This would likely not be reactive (you won't get rerenders in parts that depend on this, when `disabled` changed). Also it's basically the inverse of the paradigm of all modern JS Apps, with the DOM being a function of state, and not vice versa. 

> What do you feel is best here knowing above? I'm still on the fence! Could go either way.

Let's keep it as it is then! We can still refactor when it becomes clear we won't need it.

ember-toucan-core/rollup.config.mjs
```diff
@@ -24,7 +24,11 @@ export default {
     // These are the modules that should get reexported into the traditional
     // "app" tree. Things in here should also be in publicEntrypoints above, but
     // not everything in publicEntrypoints necessarily needs to go here.
-    addon.appReexports(['components/*.js', 'components/form/*.js']),
+    addon.appReexports([
+      'components/*.js',
+      'components/form/*.js',
+      'components/form/controls/*.js',
+    ]),
```
ynotdraw, line 31: Open to suggestions here - I played around with glob patterns but couldn't get _exactly_ what I wanted.  Essentially I wanted to ignore `components/form/-private` but include everything else, but everything I tried using the `!` pattern kept giving me not what I wanted.  This is a bit verbose 🤷 .  100% open to help/suggestions here!

ember-toucan-core/src/components/form/controls/textarea.ts
```diff
@@ -0,0 +1,18 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export interface ToucanFormTextareaControlComponentSignature {
+  Element: HTMLTextAreaElement;
+  Args: {
+    isDisabled?: boolean;
+    onChange?: (value: string, e: Event | InputEvent) => void;
```
ynotdraw, line 8: I updated this a little bit based on https://github.com/CrowdStrike/ember-toucan-core/pull/36/files#r1114409957. I personally like having access directly to the event when needed, so adding it as a second argument felt like a potential nice thing to have.  Some existing code is doing this as well, FWIW.  For example, one case was not just wanting the `value` but the `key`/`keyCode` - I could access it directly from `e`.  Totally open to suggestions here though as well!

ember-toucan-core/src/components/form/textarea-field.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { ToucanFormTextareaControlComponentSignature } from './controls/textarea';
+
+export interface ToucanFormTextareaFieldComponentSignature {
+  Element: HTMLTextAreaElement;
+  Args: {
+    error?: string;
+    hint?: string;
+    isDisabled?: boolean;
+    label: string;
+    onChange?: ToucanFormTextareaControlComponentSignature['Args']['onChange'];
```
ynotdraw, line 13: I was waffling back and forth on doing something like this in `controls/textarea`:

```js
export interface ToucanFormTextareaControlArguments {
   isDisabled?: boolean;
   onChange?: (value: string, e: Event | InputEvent) => void;
   value?: string;
 }
```

and then:

```js
onChange?: ToucanFormTextareaControlArguments['onChange'];
```

but it felt like an extra interface for not much gain.  Due to that, I opted for this as it's valid TS as well.  I have 0 strong opinions here though so will gladly do whatever folks think is best.

(My revert from above to this 😂 : https://github.com/CrowdStrike/ember-toucan-core/pull/36/commits/dc7902aff9d5429c288125a491b71e4abd71e0dd#diff-55352b7dd59af1213309f0921bd79cc8712b6d051513d50f3f23e3209ed52f11L4)

ember-toucan-core/src/components/form/textarea-field.hbs
ynotdraw: Just met with the team - we are going to wait until David is back on Monday to decide how we want to handle this.  I went ahead and filed https://github.com/CrowdStrike/ember-toucan-core/issues/44 for tracking and will pick up as soon as we get an answer!  It sounded as though we want to support what you described though - just need to figure out how it should look

ember-toucan-core/src/components/form/textarea-field.hbs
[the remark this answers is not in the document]
nicolechung: This is an example of the test helpers provided in ember-headless-table: https://ember-headless-table.pages.dev/docs/testing

ynotdraw:
Apologies - not quite sure what happened with doubling up on commits 😅.  I'll maybe squash commits on merge.

docs/components/textarea-field/index.md
```diff
@@ -0,0 +1,111 @@
...
+
+Provide a string to `@hint` to render the text into the Hint section of the Field. This is optional.
+
+## Error
+
+Provide a string to `@error` to render the text into the Error section of the Field. This is optional.
+
+## Value and onChange
+
+To tie into the input event, provide `@onChange`. `@onChange` will return two arguments, the first being the value, while the second being the raw event object. It's most common to use this in combination with `@value` which will set the value for the textarea based on the input received from the change event.
```
nicolechung, line 19: Nit, suggestion: I would bullet `value` and the `raw event object`, something like:

```
 `@onChange` will return two arguments:
 - the value of the input
 - the raw event object
```

This is because it gets buried in the middle of the paragraph and some folks are unlikely to read this.

test-app/tests/integration/components/textarea-field-test.gts
```diff
@@ -0,0 +1,181 @@
...
+      <TextareaField @label="Label" @error="Error text" data-textarea />
+    </template>);
+
+    let error = find('[data-error]');
+
+    assert.dom(error).hasText('Error text');
+    assert.dom(error).hasAttribute('id');
+
+    let errorId = error?.getAttribute('id') || '';
+    assert.ok(errorId, 'Expected error ID to be truthy');
```
nicolechung, line 69: `error ID` -> `errorId`

test-app/tests/integration/components/textarea-field-test.gts
```diff
@@ -0,0 +1,181 @@
...
+      <TextareaField
+        @label="Label"
+        @error="Error text"
+        @hint="Hint text"
+        data-textarea
+      />
+    </template>);
+
+    let errorId = find('[data-error]')?.getAttribute('id') || '';
+    assert.ok(errorId, 'Expected error ID to be truthy');
```
nicolechung, line 92: `error ID` -> `errorId`

test-app/tests/integration/components/textarea-field-test.gts
```diff
@@ -0,0 +1,181 @@
...
+
+    let hintId = find('[data-hint]')?.getAttribute('id') || '';
+    assert.ok(hintId, 'Expected hint ID to be truthy');
```
nicolechung, line 95: `hint ID` -> `hintId`

I think either ALL the descriptions should be `error ID` or `errorId` I mean....(one format or another)

test-app/tests/integration/components/textarea-field-test.gts
```diff
@@ -0,0 +1,181 @@
...
+      assert.step('handleChange');
+    };
+
+    await render(<template>
+      <TextareaField @label="Label" @onChange={{handleChange}} data-textarea />
+    </template>);
+
+    assert.verifySteps([]);
+
+    await fillIn('[data-textarea]', 'test');
```
nicolechung, line 149: Suggestion: should there be an `assert` on the textArea `value` too?

nicolechung (approved):
💯 Excited to see this coming together!

ember-toucan-core/src/components/form/controls/textarea.ts
```diff
@@ -0,0 +1,18 @@
...
+    value?: string;
+  };
+}
+
+export default class ToucanFormTextareaControlComponent extends Component<ToucanFormTextareaControlComponentSignature> {
+  @action
+  handleInput(e: Event | InputEvent): void {
+    this.args.onChange?.((e.target as HTMLTextAreaElement)?.value, e);
```
nicolechung, line 16: instead of having `as` you can assert on the element:

```
    assert('Expected HTMLTextAreaElement', e.target instanceof HTMLTextareaElement);
```

docs/components/textarea-field/index.md
ynotdraw: Thanks for the suggestion! Applied!

ember-toucan-core/src/components/form/controls/textarea.ts
ynotdraw: Interesting!  Thanks - went ahead and applied!

test-app/tests/integration/components/textarea-field-test.gts
ynotdraw: Yeah that makes sense.  I more think of these as the hint's ID, rather than being exactly tied to the variable name of `hintId`, but I definitely see your point! Went ahead and updated!

test-app/tests/integration/components/textarea-field-test.gts
ynotdraw: I think that makes sense! Went ahead and added

ember-toucan-core/rollup.config.mjs
simonihmig: In headless-form I have `-private` as a direct child of `src`, so `-private/components/form` instead of `components/form/-private`, see https://github.com/CrowdStrike/ember-headless-form/tree/main/packages/ember-headless-form/src. It's a matter of taste I guess, so not strongly advocating for that file layout. But it would make the glob simple here, as `components/**/*.js` would cover only public components then.

ember-toucan-core/src/components/form/textarea-field.ts
simonihmig: I think this will be a common function signature, which e.g. the input component (control and -field) would also  need, right? In which case we could define it in a shared `types.ts`, something like:

```ts
type OnChangeCallback = (value: string, e: Event | InputEvent) => void;
// or, if we want to support values other than strings:
type OnChangeCallback<T> = (value: T, e: Event | InputEvent) => void;
```

I have similar shared types here: https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/types.ts#L35-L38

ember-toucan-core/src/components/form/textarea-field.hbs
```diff
@@ -0,0 +1,28 @@
+<div
+  class="flex flex-col"
+  data-root-field={{if @rootTestSelector @rootTestSelector null}}
```
simonihmig, line 3: Nit: is `null` required here?

simonihmig (approved):
Great work!

You may want to look into some of my recent comments, and feel free to address or ignore, as there is nothing blocking this from merging IMO.

ember-toucan-core/src/components/form/textarea-field.hbs
ynotdraw: Nope! Went ahead and removed 🙇

ember-toucan-core/rollup.config.mjs
ynotdraw: I like this idea! It also helps us be consistent between the two repos, which I like.  I went ahead and followed the same pattern as headless-form and will push the change in just a bit! Thanks for the suggestion!  I also like how the glob here is way simpler like you said with `components/**/*.js` only

ember-toucan-core/src/components/form/textarea-field.ts
ynotdraw: Love this idea.  Went ahead and added it to a shared `types.ts` as suggested!

PR #42, opened 2023-02-21
#41: Add Tailwind prettier plugin+run it

### LINKED ISSUE #41: Add prettier-plugin-tailwindcss
It'd be nice to use https://tailwindcss.com/blog/automatic-class-sorting-with-prettier internally so we have less opinions on class name sorting.

ynotdraw:
## 💅  Description

This PR adds `prettier-plugin-tailwindcss` to the docs-app + `ember-toucan-core` package.

Closes #41 

---

## 📚 Documentation

- https://tailwindcss.com/blog/automatic-class-sorting-with-prettier

---

## 🔬 How to Test

- Non-visual change, only code formatting so builds need to be green.

---

N/A

### DIFF
--- a/docs-app/.prettierrc.js
+++ b/docs-app/.prettierrc.js
@@ -11,6 +11,7 @@ module.exports = {
       files: '*.hbs',
       options: {
         singleQuote: false,
+        plugins: ['prettier-plugin-tailwindcss'],
       },
     },
     {
--- a/ember-toucan-core/package.json
+++ b/ember-toucan-core/package.json
@@ -96,6 +96,7 @@
     "postcss": "^8.2.14",
     "prettier": "^2.8.3",
     "prettier-plugin-ember-template-tag": "^0.3.0",
+    "prettier-plugin-tailwindcss": "^0.2.3",
     "rollup": "^3.12.1",
     "rollup-plugin-copy": "^3.4.0",
     "rollup-plugin-ts": "^3.0.2",
--- a/ember-toucan-core/src/components/form/-private/error.hbs
+++ b/ember-toucan-core/src/components/form/-private/error.hbs
@@ -1,4 +1,4 @@
-<div class="type-xs-tight text-critical flex items-center mt-1" ...attributes>
+<div class="type-xs-tight text-critical mt-1 flex items-center" ...attributes>
   {{! Icon taken from Tony's open source icon set, this is temporary! }}
   <svg
     xmlns="http://www.w3.org/2000/svg"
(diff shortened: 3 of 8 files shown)


docs-app/.prettierrc.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
       files: '*.hbs',
```
ynotdraw, line 11: Only added to `*.hbs` files for now as it doesn't seem to work yet with `.gts/gjs`

ember-toucan-core/package.json
```diff
@@ -96,6 +96,7 @@
     "postcss": "^8.2.14",
     "prettier": "^2.8.3",
     "prettier-plugin-ember-template-tag": "^0.3.0",
+    "prettier-plugin-tailwindcss": "^0.2.3",
```
nicolechung, line 99: Question: what's the difference between this and https://www.npmjs.com/package/ember-template-lint-plugin-tailwindcss?

ember-toucan-core/src/components/form/-private/error.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="type-xs-tight text-critical flex items-center mt-1" ...attributes>
+<div class="type-xs-tight text-critical mt-1 flex items-center" ...attributes>
```
nicolechung, line 1: Looks like it's reordering! 🎉  this is GREAT!

nicolechung (approved):

ember-toucan-core/package.json
ynotdraw: Good question! I'm not entirely sure to be honest 😅!  Poking around a little bit, my _best guess_ would be what you linked was maybe created either before the **official** `prettier-plugin-tailwindcss` was released or to be specifically used with the ember-template-lint plugin system (or maybe both?).

FWIW, it doesn't look like that package uses `prettier-plugin-tailwindcss` under the hood either.

The way our prettier rules work today is by using `overrides` and target `.hbs/.gts/.gjs` extensions rather than adding a template-lint directly it appears via https://github.com/CrowdStrike/ember-toucan-core/blob/main/ember-toucan-core/.prettierrc.cjs#L11.  This is pretty interesting!  So, as another guess, I'd imagine by leveraging `overrides` like this, we get to use the official tooling from the Tailwind folks (yay!) without having to be specific to Ember/Ember's template linting.

I apologize I don't have a better answer other than pure speculation (😂)!

simonihmig (approved):
Nice!

PR #50, opened 2023-02-23
#48: Add Checkbox component

### LINKED ISSUE #48: Create Checkbox component
Depends on https://github.com/CrowdStrike/ember-toucan-core/issues/10

As a developer, I'd like to have a CheckboxField and Checkbox Component, so that I may have standardized styles without styling the component myself.  It should (likely?) build on top of `<Field` introduced in the above issue.

**Acceptance criteria**
- Checkbox has a label argument
- Checkbox has a hint argument that can be displayed optionally.
- Checkbox has a control argument that can accept different input type="checkbox" attributes
- Checkbox has an error argument
- Error text is only displayed during errors

ynotdraw:
## 🚀 Description
This PR introduces a `Checkbox` control component and `CheckboxField` component. 

**NOTE:** These components are not 100% styled yet, as we are still figuring out where to land on a few things. This is about ~75% close though.

Closes #48 

---

## 🔬 How to Test

- These are only presentational components, so test should pass
- View the checkbox link at https://48034b83.ember-toucan-core.pages.dev/docs/components/checkbox and review the documentation
  - **NOTE:** This checkbox does not toggle on click! It is presentational only
- View the link at https://48034b83.ember-toucan-core.pages.dev/docs/components/checkbox-field  and review the documentation
  - **NOTE:** These components are not 100% styled yet, as we are still figuring out where to land on a few things.  This is about ~75% close though.

---

**UI States**
Screenshot 2023-02-24 at 10 12 02 AM

**Focus State**

https://user-images.githubusercontent.com/8069555/221663299-f39a228d-c505-4bca-abfb-3f7dddc25ce2.mov

**A11y**

Showing the screenreader picking up:

1. Checkbox state
2. aria-invalid
3. hint text
4. error text
5. hint + error text

https://user-images.githubusercontent.com/8069555/221663069-85e1efe8-6add-4380-8e5f-c169bca8804f.mov

### DIFF
--- a/docs/components/checkbox-field/demo/base-demo.md
+++ b/docs/components/checkbox-field/demo/base-demo.md
@@ -13,7 +13,6 @@ import { tracked } from '@glimmer/tracking';
 
 export default class extends Component {
   @tracked value = false;
-  @tracked errorMessage;
 
   @action
   updateValue(value) {
--- a/docs/components/checkbox-field/demo/base-demo.md
+++ b/docs/components/checkbox-field/demo/base-demo.md
@@ -1,0 +1,23 @@
+```hbs template
+<Form::CheckboxField
+  @label='Label'
+  @value={{this.value}}
+  @onChange={{this.updateValue}}
+/>
+```
+
+```js component
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class extends Component {
+  @tracked value = false;
+  @tracked errorMessage;
+
+  @action
+  updateValue(value) {
+    this.value = value;
+  }
+}
+```
--- a/docs/components/checkbox-field/index.md
+++ b/docs/components/checkbox-field/index.md
@@ -16,13 +16,14 @@ Provide a string to `@error` to render the text into the Error section of the Fi
 
 ## Value and onChange
 
-To tie into the input event, provide `@onChange`. `@onChange` will return three arguments:
+To tie into the input event, provide `@onChange`. `@onChange` will return two arguments:
 
 1. the checked attribute from the target
 2. the raw event object
-3. the indeterminate property from the target
 
 It's most common to use this in combination with `@value` which will set the `checked` attribute for the Checkbox.
+
+To access the indeterminate property of the checkbox, use `e.target.indeterminate`.
 
 ```hbs
 <Form::CheckboxField
@@ -41,8 +42,8 @@ export default class extends Component {
   @tracked value;
 
   @action
-  handleChange(value, e, indeterminate) {
-    console.log({ e, indeterminate, value });
+  handleChange(value, e) {
+    console.log({ e, indeterminate: e.target.indeterminate, value });
     this.value = value;
   }
 }
--- a/docs/components/checkbox-field/index.md
+++ b/docs/components/checkbox-field/index.md
@@ -20,7 +20,7 @@ To tie into the input event, provide `@onChange`. `@onChange` will return three 
 
 1. the checked attribute from the target
 2. the raw event object
-3. the indeterminate attribute from the target
+3. the indeterminate property from the target
 
 It's most common to use this in combination with `@value` which will set the `checked` attribute for the Checkbox.
 
--- a/docs/components/checkbox-field/index.md
+++ b/docs/components/checkbox-field/index.md
@@ -1,0 +1,158 @@
+# Checkbox Field
+
+Provides an underlying checkbox element building on top of the Field component.
+
+## Label
+
+Provide a string to `@label` to render the text into the `<label>` of the Field.
+
+## Hint
+
+Provide a string to `@hint` to render the text into the Hint section of the Field. This is optional.
+
+## Error
+
+Provide a string to `@error` to render the text into the Error section of the Field. This is optional.
+
+## Value and onChange
+
+To tie into the input event, provide `@onChange`. `@onChange` will return three arguments:
+
+1. the checked attribute from the target
+2. the raw event object
+3. the indeterminate attribute from the target
+
+It's most common to use this in combination with `@value` which will set the `checked` attribute for the Checkbox.
+
+```hbs
+<Form::CheckboxField
+  @label='Label'
+  @value={{this.value}}
+  @onChange={{this.handleChange}}
+/>
+```
+
+```js
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class extends Component {
+  @tracked value;
+
+  @action
+  handleChange(value, e, indeterminate) {
+    console.log({ e, indeterminate, value });
+    this.value = value;
+  }
+}
+```
+
+## Disabled State
+
+Set the `@isDisabled` argument to disable the checkbox.
+
+## Indeterminate State
+
+Set the `@isIndeterminate` argument. To learn more, view the documentation Checkbox documentation.
+
+## Attributes and Modifiers
+
+Consumers have direct access to the underlying [checkbox element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox), so all attributes are supported. Modifiers can also be added directly to the Checkbox as shown in the demo.
+
+## Test Selectors
+
+### Root Element
+
+Provide a custom selector via `@rootTestSelector`. This test selector will be used as the value for the `data-root-field` attribute. The Field can be targeted via:
+
+```hbs
+<Form::CheckboxField @label='Label' @rootTestSelector='example' />
+```
+
+```js
+assert.dom('[data-root-field="example"]');
+
+// targeting this field's specific label
+assert.dom('[data-root-field="example"] > [data-label]');
+```
+
+### Label
+
+Target the label element via `data-label`.
+
+### Hint
+
+Target the hint block via `data-hint`.
+
+### Error
+
+Target the error block via `data-error`.
+
+## All UI States
+
+<div class="flex flex-col space-y-4" style="max-width: 14rem">
+<Form::CheckboxField
+@label='Label'
+/>
+
+<Form::CheckboxField
+@label='Label'
+@error="With error"
+/>
+
+<Form::CheckboxField
+@label='Label'
+@hint='With hint text'
+/>
+
+<Form::CheckboxField
+@label='Label'
+@hint='With hint text'
+@error="With error"
+/>
+
+<Form::CheckboxField
+@label='Label'
+@hint='Checked'
+@value={{true}}
+/>
+
+<Form::CheckboxField
+@label='Label'
+@hint='Indeterminate'
+@isIndeterminate={{true}}
+/>
+
+<Form::CheckboxField
+@label='This is an option that expands to multiple lines'
+/>
+
+<Form::CheckboxField
+@label='This is an option that expands to multiple lines'
+@hint="Here is helper text that overflows onto multiple lines"
+/>
+
+<Form::CheckboxField
+@label='This is an option that expands to multiple lines'
+@error="With error"
+/>
+
+<Form::CheckboxField
+@label='Disabled'
+@isDisabled={{true}}
+/>
+
+<Form::CheckboxField
+@label='Disabled'
+@hint="With hint"
+@isDisabled={{true}}
+/>
+
+<Form::CheckboxField
+@label='Disabled + checked'
+@value={{true}}
+@isDisabled={{true}}
+/>
+
+</div>
--- a/docs/components/checkbox-field/index.md
+++ b/docs/components/checkbox-field/index.md
@@ -54,7 +54,7 @@ Set the `@isDisabled` argument to disable the checkbox.
 
 ## Indeterminate State
 
-Set the `@isIndeterminate` argument. To learn more, view the documentation Checkbox documentation.
+Set the `@isIndeterminate` argument. To learn more, view the [Checkbox documentation](./checkbox).
 
 ## Attributes and Modifiers
 
--- a/docs/components/checkbox-field/index.md
+++ b/docs/components/checkbox-field/index.md
@@ -89,6 +89,10 @@ Target the hint block via `data-hint`.
 
 Target the error block via `data-error`.
 
+### Control
+
+The checkbox control is wrapped in a `<div>` so that an error box shadow can be added. Target the control block via `data-control`.
+
 ## All UI States
 
 <div class="flex flex-col space-y-4" style="max-width: 14rem">
--- a/docs/components/checkbox-field/index.md
+++ b/docs/components/checkbox-field/index.md
@@ -89,10 +89,6 @@ Target the hint block via `data-hint`.
 
 Target the error block via `data-error`.
 
-### Control
-
-The checkbox control is wrapped in a `<div>` so that an error box shadow can be added. Target the control block via `data-control`.
-
 ## All UI States
 
 <div class="flex flex-col space-y-4" style="max-width: 14rem">
--- a/docs/components/checkbox/index.md
+++ b/docs/components/checkbox/index.md
@@ -7,19 +7,29 @@ Provides a Toucan-styled [checkbox element](https://developer.mozilla.org/en-US/
 To set the `checked` attribute of the checkbox, provide `@value`.
 
 ```hbs
-<Form::Controls::Checkbox @value={{true}} />
-<Form::Controls::Checkbox @value={{false}} />
+<Form::Controls::Checkbox @value={{true}} data-checkbox-1 />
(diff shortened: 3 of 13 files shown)


ember-toucan-core/src/components/form/checkbox-field.hbs
```diff
@@ -0,0 +1,41 @@
...
+  class="flex flex-col"
+  data-root-field={{if @rootTestSelector @rootTestSelector}}
+>
+  <Form::Field as |field|>
+    <field.Control
+      class="rounded-sm p-1 {{if @error 'shadow-error-outline'}}"
+      data-control
+    >
+
+      <label class="flex flex-row space-x-3" for={{field.id}}>
```
ynotdraw, line 11: Can't use field's yielded label as we need different leading as well as use a different text color for checkbox labels. I verified with Mary and David this is the expected look and that we don't want to use the styles from field.Label

ember-toucan-core/src/components/form/checkbox-field.hbs
```diff
@@ -0,0 +1,41 @@
...
+        <Form::Controls::Checkbox
+          aria-invalid={{if @error "true"}}
+          aria-describedby="{{if @error field.errorId}}"
```
ynotdraw, line 14: Since the label text (span below) + hint are wrapped inside the label, we don't need to add them to `aria-describedby` here for screenreaders 🎉

ember-toucan-core/src/components/form/controls/checkbox.ts
```diff
@@ -0,0 +1,74 @@
...
+    let isCheckedOrIndeterminate = this.isCheckedOrIndeterminate;
+
+    return isCheckedOrIndeterminate && !isDisabled
+      ? ['bg-primary-idle border-none']
+      : isCheckedOrIndeterminate && isDisabled
+      ? ['bg-disabled border-none']
+      : !isCheckedOrIndeterminate && isDisabled
+      ? ['bg-transparent border-disabled']
+      : ['bg-normal-idle'];
+  }
```
ynotdraw, line 66: These are super specific styles based on state, so I wrote assertions in tests for them 😅

docs/components/checkbox-field/demo/base-demo.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+```js component
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class extends Component {
+  @tracked value = false;
+  @tracked errorMessage;
```
nicolechung, line 16: Suggestion: remove this errorMessage...or use it, since it's not being used.

ember-toucan-core/unpublished-development-types/index.d.ts
```diff
@@ -3,13 +3,10 @@
 
 import '@glint/environment-ember-loose';
 
-import type TextareaControlComponent from '../src/components/form/controls/textarea';
-import type FieldComponent from '../src/components/form/field';
+import type ToucanCoreRegistry from '../src/template-registry';
 
 declare module '@glint/environment-ember-loose/registry' {
-  export default interface Registry /* extends EmberPageTitle, etc, other addon registries */ {
+  export default interface Registry extends ToucanCoreRegistry {
```
nicolechung, line 9: Nice! Simplifies things for sure

test-app/tests/integration/components/checkbox-field-test.gts
```diff
@@ -0,0 +1,223 @@
...
+
+    assert.dom('[data-checkbox]').hasAttribute('aria-invalid', 'true');
+
+    assert
+      .dom('[data-root-field="test"] > [data-control]')
+      .hasClass('shadow-error-outline');
+  });
+
+  test('it sets the "for" attribute on the label to the "id" attribute of the checkbox', async function (assert) {
+    await render(<template>
```
nicolechung, line 87: I really like how small these tests are...makes it easier to read for sure.

test-app/tests/integration/components/checkbox-field-test.gts
```diff
@@ -0,0 +1,223 @@
...
+    assert.expect(4);
+
+    let handleChange = (
+      _checked: boolean,
+      _e: Event | InputEvent,
+      isIndeterminate: boolean
+    ) => {
+      assert.false(
+        isIndeterminate,
+        'Expected indeterminate state to be false as we did not provide `@isIndeterminate={{true}}`'
```
nicolechung, line 175: Didn't we provide `@isIndeterminate={{true}}` on line 184 (below)? Does clicking make `Indeterminate` false?

nicolechung: I always thought `isIndeteminate` is for bulk selection...

docs/components/checkbox-field/index.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+## Disabled State
+
+Set the `@isDisabled` argument to disable the checkbox.
+
+## Indeterminate State
+
+Set the `@isIndeterminate` argument. To learn more, view the documentation Checkbox documentation.
```
danwenzel, line 57:
```suggestion
Set the `@isIndeterminate` argument. To learn more, view the Checkbox documentation.
```

Could we also have it be a hyperlink to that documentation?

docs/components/checkbox-field/index.md
```diff
@@ -0,0 +1,158 @@
...
+Target the hint block via `data-hint`.
+
+### Error
+
+Target the error block via `data-error`.
+
+## All UI States
+
+<div class="flex flex-col space-y-4" style="max-width: 14rem">
+<Form::CheckboxField
```
danwenzel, line 95: can the indentation be fixed on all these, or is this a limitation of the `md` files?

docs/components/checkbox/index.md
```diff
@@ -0,0 +1,55 @@
...
+To set the `checked` attribute of the checkbox, provide `@value`.
+
+```hbs
+<Form::Controls::Checkbox @value={{true}} />
+<Form::Controls::Checkbox @value={{false}} />
+```
+
+## Indeterminate
+
+Checkboxes have the ability to be in the [indeterminate state](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes). This is accomplished with `@isIndeterminate`. It is up to the consumer to decide what will force the checkbox into the indeterminate state. The [data-checked](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked) attribute is set to either "true", "false", or "mixed" depending on the state (where "mixed" is set if `@isIndeterminate={{true}}`).
```
danwenzel, line 16: It's a bit confusing to me that all of the `aria-checked` attribute values are mentioned here, in the "Indeterminate" section.  Can we mention only the `mixed` one here, and then mention `true` and `false` up in the "Checked Value" section?

docs/components/checkbox/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+```hbs
+<Form::Controls::Checkbox @isIndeterminate={{true}} />
+```
+
+## onChange
+
+To tie into the input event, provide `@onChange`. `@onChange` will return three arguments:
+
+1. the checked attribute from the target
```
danwenzel, line 26: Would this be the checked state before or after the change? Might be worth mentioning that here

ember-toucan-core/src/components/form/checkbox-field.hbs
```diff
@@ -0,0 +1,41 @@
...
+
+        <div class="flex flex-col">
+          <span
+            class="type-md-tight block leading-4
+              {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
+            data-label
+          >{{@label}}</span>
+
+          {{#if @hint}}
+            <field.Hint data-hint>{{@hint}}</field.Hint>
```
danwenzel, line 31: What is `data-hint` for, and could it be instead defined inside `field.Hint`? (same for `data-error` below)

test-app/tests/integration/components/checkbox-field-test.gts
```diff
@@ -0,0 +1,223 @@
+/* eslint-disable no-undef -- Until https://github.com/ember-cli/eslint-plugin-ember/issues/1747 is resolved... */
+/* eslint-disable simple-import-sort/imports,padding-line-between-statements,decorator-position/decorator-position -- Can't fix these manually, without --fix working in .gts */
+
+import { click, find, render, setupOnerror } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+
+import CheckboxField from '@crowdstrike/ember-toucan-core/components/form/checkbox-field';
+import { setupRenderingTest } from 'test-app/tests/helpers';
+
+module('Integration | Component | CheckboxField', function (hooks) {
```
danwenzel, line 10: Great tests!

danwenzel (approved):

test-app/tests/integration/components/checkbox-field-test.gts
simonihmig: I agree it's a bit confusing. I believe what is happening is that by clicking the checkbox, the browser itself sets `element.indeterminate` to false. And `@onChange` is _not_ called with what is passed as `@isIndeterminate`, but with the actual property value of the input element (which again has changed). That's why we receive false, while still true is passed as `@isIndeterminate`. Am I understanding this right?

But then the assertion message is wrong. Should it be something like this?

```suggestion
        'Expected indeterminate state to be false as the indeterminate property on the input element was changed by clicking it`'
```

test-app/tests/integration/components/checkbox-test.gts
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  test('it makes the checkbox indeterminate using `@isIndeterminate`', async function (assert) {
+    await render(<template>
+      {{! we do not require a label, but instead suggest using Field / TextareaField }}
+      {{! template-lint-disable require-input-label }}
+      <CheckboxControl @isIndeterminate={{true}} data-checkbox />
+    </template>);
+
+    assert.dom('[data-checkbox]').hasAttribute('data-checked', 'mixed');
```
simonihmig, line 63:
```suggestion
    assert.dom('[data-checkbox]').hasAttribute('data-checked', 'mixed');
    assert.dom('[data-checkbox]').hasProperty('indeterminate', true);
```

ember-toucan-core/src/components/form/controls/checkbox.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * We use the data-checked attribute to help signal what state the checkbox is in.
+   * This is most helpful in tests. The "mixed" value is used for the indeterminate state.
+   * We don't use aria-checked directly, as we are a valid <input type="checkbox", but
+   * we follow the possible states outlined by aria-checked.
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked
+   */
+  get checkedState() {
```
simonihmig, line 43: I am not entirely sure about why we need this? 

You could check all possible states in tests using `isChecked()` or `hasProperty('indeterminate', true)`, right?

ember-toucan-core/src/components/form/controls/checkbox.ts
```diff
@@ -0,0 +1,74 @@
...
+import { action } from '@ember/object';
+
+export interface ToucanFormCheckboxControlComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    isDisabled?: boolean;
+    onChange?: (
+      isChecked: boolean,
+      e: Event | InputEvent,
+      isIndeterminate: boolean
```
simonihmig, line 12: Would this be ever true? Like when you click an indeterminate checkbox, it will get either checked or unchecked, but the `indeterminate` property will always turn to false by the browser, won't it?

Also, this status is a bit redundant, as we already have the event, so a user could just ask for `e.target.indeterminate`. Same goes for `isChecked` of course, but that is what people will want to use in 99,9% of times, so it's reasonable to pass that as the primary value of `onChange`. But for `isIndeterminate` I am thinking whether we can drop it here from the list of agruments? 🤔

simonihmig (commented):
Great work! Only minor comments and suggestions...

docs/components/checkbox-field/index.md
ynotdraw: Yeah unfortunately it's a limitation with these files for the time being.  We could turn off prettier for markdown files to retrain the indentation, but then that could lead to other issues :(

docs/components/checkbox/index.md
ynotdraw: I like it! Updating now, thank you!

docs/components/checkbox/index.md
ynotdraw: Sure thing - thank you!

ember-toucan-core/src/components/form/checkbox-field.hbs
ynotdraw: Yeah mostly a test-selector for getting to that block in tests.  We may move those internally to `<Field>` if we keep using this as a pattern.  I'll file a task!

test-app/tests/integration/components/checkbox-field-test.gts
```diff
@@ -0,0 +1,223 @@
...
+  });
+
+  test('it renders with a hint', async function (assert) {
+    await render(<template>
+      <CheckboxField @label="Label" @hint="Hint text" data-checkbox />
+    </template>);
+
+    let hint = find('[data-hint]');
+
+    assert.dom(hint).hasText('Hint text');
```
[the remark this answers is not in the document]
ynotdraw: Definitely not!  Just copy pasta I think from the Field component.  I was using `find()` whenever I was also grabbing things like attributes off the element, but in this case here it can definitely use qunit-dom instead!  Thank you for catching this!

test-app/tests/integration/components/checkbox-field-test.gts
```diff
@@ -0,0 +1,223 @@
...
+
+    let labelFor = find('[data-control] > label')?.getAttribute('for') || '';
+    assert.ok(labelFor, 'Expected the id attribute of the label to be truthy');
+
+    let checkboxId = find('[data-checkbox]')?.getAttribute('id') || '';
+    assert.strictEqual(
+      checkboxId,
+      labelFor,
+      'Expected the for attribute on the label to match the id attribute on the checkbox'
+    );
```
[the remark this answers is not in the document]
ynotdraw: Excellent catch, thank you!

ember-toucan-core/src/components/form/controls/checkbox.ts
ynotdraw: Yeah maybe that makes more sense.  What do you think?  My initial reaction was to provide something to make it easy to check states in tests, but maybe instead we write some documentation suggesting what you have here instead?

ember-toucan-core/src/components/form/controls/checkbox.ts
ynotdraw: Yeah, maybe so! I was looking at some existing implementation and that's why I added this here, but you're right.  Maybe we just remove it altogether and if folks need to look at `indeterminate` they can pull it off of `e.target` as you mention.  That may make things a bit easier overall.

docs/components/checkbox-field/demo/base-demo.md
ynotdraw: Thanks for catching this! Applied your feedback

docs/components/checkbox-field/index.md
ynotdraw: Good idea! Added!

ember-toucan-core/src/components/form/controls/checkbox.ts
```diff
@@ -0,0 +1,74 @@
...
+      : !isCheckedOrIndeterminate && isDisabled
+      ? ['bg-transparent border-disabled']
+      : ['bg-normal-idle'];
+  }
+
+  @action
+  handleInput(e: Event | InputEvent): void {
+    assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
+
+    this.args.onChange?.(e.target?.checked, e, e.target?.indeterminate);
```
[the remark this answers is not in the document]
ynotdraw: Yes! Thank you!

ember-toucan-core/src/components/form/controls/checkbox.ts
ynotdraw: Went ahead and went this route!

simonihmig (approved):
🚀

test-app/tests/integration/components/checkbox-field-test.gts
nicolechung: Okay thanks this clarifies this for me.

PR #64, opened 2023-03-06
55 - file upload component

nicolechung:
- 🚫  did not add an controls class,  used a regular `input` tag because quite a bit of custom styling was needed (could not use `Form::Controls::Input`
- 🚫  did not add a `@values` argument, as you can't modify the value of a `<input type=file>`, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#value
- ⚡  Looks like any change event target has `FileList`. `FileList` is getting deprecated, I believe because it is very inconvenient to use compared to an Array. `FileList` is read-only, and you can't create a new FileList by `new FileList()` (this is not allowed)

## What changed recently
- [x] added `@onChange` argument
  - [x] added internal `onChange` handler. This takes the FileList and converts to an array of File objects (`File[]`), which I then pass to the args `@onChange`. 
- [x] added @files argument which accepts and array of File objects (`File[]`).
- [x] removed `@onDelete` argument which takes the current file
- [x] update filenames in the `docs` so that multiple demos can exist under the same category. **Note**: this is a bit magical and I logged it as an issue here: https://github.com/CrowdStrike/ember-oss-docs/issues/15
- [x] created sub-components of `Form::FileInputField` and added related tests:
  - [x] delete-button
    - [x] added `aria-label`.
  - [x] list-item
  - [x] list
  - [x] file-input (this is the markup inside of the `field.Control`
 - [x] added two tests for the `multiple` attribute
   - [x] `multiple=false` does a `replace`
   - [x] `multiple=true` `adds to` what already exists in the list 

## 🧱 Component
- FileInputField
- updated template-registry and package.json
- added new types as `Event`, `InputEvent` and `MouseEvent` do not have a `files` key/value on their `EventTarget`. 

## 📚  Docs
- Added docs (using TextAreaField as a template)
- Added docs demo
- added UI states demo as I need `@onChange` and `@onDelete` which require a backing class
- 🚫  did not add a `with files` state as I think I need JS/TS for this (for creating a `FileList` array). Again, any tips would be appreciated! Would be nice to include it.

## 💥  Tests
- again, used the TextAreaField as a template for the tests
- onChange works a bit differently, the first parameter is `files` (FileList) instead of `value` (string)


ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,74 @@
...
+      args.label !== undefined
+    );
+    super(owner, args);
+
+    if (args.files) {
+      this.files = args.files;
+    }
+  }
+
+  get accept() {
```
nicolechung, line 47: Props/Arg question: I am conflicted about mixing args and @tracked values like this (in React it's considered a bit [of a code smell](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state))

This pattern does allow default @args however (for `accepts` and `multiple`). 

Is there a better way to handle default @args yet? I remember Preston saying there wasn't...

ynotdraw (commented):
HUGE kudos for taking this on - this is definitely one of the tricker components! 🎉

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+          </div>
+        </div>
+      </field.Control>
+    </field.Label>
+    {{#if @error}}
+      <field.Error id={{field.errorId}} data-error>{{@error}}</field.Error>
+    {{/if}}
+    <ul class="px-0" data-files>
+      {{#each this.files as |file|}}
+        <li class="flex items-center p-0">
```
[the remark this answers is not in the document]
nicolechung: Yeah! 

I adjust the padding but I'm a little unclear how the Figma styles for the shadow outline translates to our tailwind classes.

docs/components/file-input-field/index.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+## Disabled State
+
+Set the `@isDisabled` argument to disable the `<input type="file">`.
+
+## Attributes and Modifiers
+
+Consumers have direct access to the underlying [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea), so all attributes are supported. Modifiers can also be added directly to the input as shown in the demo.
```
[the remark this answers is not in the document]
nicolechung: I thought I caught everything! Clearly missed one 😅  - thank you for catching this!

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
+<div
+  class="flex flex-col {{if @isDisabled 'text-disabled cursor-not-allowed'}}"
+  data-root-field={{if @rootTestSelector @rootTestSelector null}}
+>
+  <Form::Field as |field|>
+    {{!-- <field.Label for={{field.id}} data-label>{{@label}}</field.Label> --}}
```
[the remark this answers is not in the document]
nicolechung: removed!

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+            multiple={{this.multiple}}
+            {{! template-lint-disable no-bare-strings no-inline-styles  }}
+            style="clip: rect(1px, 1px, 1px, 1px);"
+            ...attributes
+            {{on "change" this.handleChange}}
+          />
+          <div
+            class="text-titles-and-attributes inline-flex w-full cursor-pointer items-center pl-2"
+          >
+            {{! TODO: replace with official icon later }}
```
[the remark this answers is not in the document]
nicolechung: Yes! Thank you for the issue link!

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+            <p
+              class="text-titles-and-attributes m-0 truncate p-0"
+              data-file-name
+            >{{file.name}}</p>
+            <span
+              class="body-and-labels block"
+              data-file-size
+            >{{(this.formatSize file.size)}}</span>
+          </section>
+          <div id="trash" class="flex-none">
```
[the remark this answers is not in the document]
nicolechung: Did a few things here:
1. changed `FileList` to an Array. I left a comment about why (apparentlyFileList is getting deleted)
2. changed `div` to `button` and removed the `id` since that made no sense (there can be multiple buttons)
3. added `onDeleteFile` arg in case folks want to do something else with the deleted file
4. added a delete handler to handle the internal state.

We could move all of this external (to the consumer) but my concern in that case is users will have some boilerplate code they will often just copy over to their component. Having internal state handles this for them.

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,75 @@
...
+      event.target instanceof HTMLInputElement
+    );
+
+    if (event.target.files) {
+      this.files = event.target.files;
+
+      return this.args.onChange?.(event.target.files, event);
+    }
+
+    return this.args.onChange?.(new FileList(), event);
```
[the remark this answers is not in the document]
nicolechung: This should be switched to `File[]` now (left comments as to why).

This FileList, while iterable, is not a very useful object since it's missing a lot of `Array` methods like `filter`, `splice`.

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+            type="file"
+            accept={{this.accept}}
+            multiple={{this.multiple}}
+            {{! template-lint-disable no-bare-strings no-inline-styles  }}
+            style="clip: rect(1px, 1px, 1px, 1px);"
+            ...attributes
+            {{on "change" this.handleChange}}
+          />
+          <div
+            class="text-titles-and-attributes inline-flex w-full cursor-pointer items-center pl-2"
```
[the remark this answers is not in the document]
nicolechung: `gap-x-2` seemed to work?

ember-toucan-core/src/components/form/file-input-field.ts
simonihmig: No, I think it's fine!

I had only a quick look at your React document, and also I don't know a lot about React, but I don't see _any_ problem here (except for maybe having to type this here in a verbose way, so something for "default value for args" was [discussed](https://github.com/emberjs/rfcs/issues/479) a while ago, doesn't seem it will come tho).

> I am conflicted about mixing args and @Tracked values like this

But you are _not_ doing this! There is no `@tracked` here. The getter is fully derived state, nothing except for the actuals arg is tracked here, so also nothing can get out of sync or diverge in any way. 

It would be different if you did this instead:

```js
@tracked accept = this.args.accept ?? '*';
```

Then indeed you would have two copies of `accept`. But you are not actually overriding the arg with some local state or so. Whenever `this.arg.accept` changes, `this.accept` will automatically reflect that. In that sense, I think what we have here is basically an [controlled component](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-controlled-component) as defined in your React doc.

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,274 @@
...
+
+    let errorId = error?.getAttribute('id') || '';
+    assert.ok(errorId, 'Expected errorId to be truthy');
+
+    // For the file input field component, the only aria-describedby
+    // value should be the errorId.  This is due to the way the component
+    // is structured, where the label+hint are rendered inside of the
+    // wrapping <label> element
+    let describedby =
+      find('[data-file-input-field]')?.getAttribute('aria-describedby') || '';
```
simonihmig, line 100: nitpick, but in general I would always use `??` instead of `||` for applying a default value. In this case it does not matter, but for things that e.g. return a boolean it is tricky, as a legit value could be falsey then and overridden by the default value. Like `return this.isValid || true`, this would _always_ be true, even when `this.isValid` returns `false`. But `??` would do the expected thing here. And so for consistency, I would just think of `||` as a boolean or operation, and `??` as "apply a default value". 

Just my way of thinking about this, but feel free to ignore! 😏

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,274 @@
...
+    assert.dom('[data-files] [data-file-size]').hasText('0 KB');
+  });
+
+  test('it calls @onChange when input is received', async function (assert) {
+    assert.expect(6);
+
+    const handleChange = (files: File[], e: Event | InputEvent) => {
+      assert.ok(e, 'Expected `e` to be available as the second argument');
+      assert.ok(e.target, 'Expected direct access to target from `e`');
+      assert.strictEqual(files.length, 1, 'Expected a single file to be uploaded');
```
simonihmig, line 186: Can we check more that just that this is an array? Like `files[0] instanceof File`, and maybe the name of it?

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,85 @@
...
+    onDeleteFile?: (file: File) => void;
+    rootTestSelector?: string;
+  };
+  Blocks: {
+    triggerText: [];
+  };
+}
+
+export default class ToucanFormFileInputFieldComponent extends Component<ToucanFormFileInputFieldComponentSignature> {
+  @tracked files: File[] = [];
```
simonihmig, line 32: 1. nit: you _could_ have assigned the arg here also, instead of separately in the constructor below, like 
```suggestion
  @tracked files: File[] = this.args.files ?? [];
```
2. so relating to the discussion below on "mixing args and tracked state", this is indeed a different case than `accept` and the other getters, and I wonder if that is really what we want here.
We could instead also treat this component as "controlled" in the react sense, and not have any local state. Which means 
  * we only look at the `@files` arg, even for rendering the files list 
  * `onChange` the user would be responsible to update the file list, which is state that _they own_, and let it flow back into `@files`

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,85 @@
...
+  Args: {
+    accept?: string;
+    error?: string;
+    files?: File[];
+    hint?: string;
+    label: string;
+    isDisabled?: boolean;
+    multiple?: boolean;
+    onChange?: OnFileUploadChangeCallback;
+    onDeleteFile?: (file: File) => void;
```
simonihmig, line 23: I wonder about the API design of these two callbacks here.
* when a file is deleted, a user could maybe expect `@onChange` to be called also, as the file list _has_ changed. But this does not seem to happen, right?
* we could also have just `@onChange`, and call that with the now smaller list of files when deleting a file
* or we rename `@onChange` to `@onAddFile` or something like this, to make this clear, when we want to have different change callbacks

🤔

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+              viewBox="0 0 24 24"
+            ><path
+                d="M3.4 6.2h17.2m-15.3 0V19a2.063 2.063 0 00.538 1.413A1.836 1.836 0 007.2 21h9.6a1.861 1.861 0 001.325-.587A2.039 2.039 0 0018.7 19V6.2M10 15.6v-4m4 0v4M9.4 6.2V3.9a.98.98 0 01.212-.637A.746.746 0 0110.2 3h3.6a.743.743 0 01.587.263.975.975 0 01.213.637v2.3"
+                fill="none"
+                stroke-linecap="round"
+                stroke-linejoin="round"
+                stroke-width="2"
+              ></path></svg>
+            {{#if (has-block "triggerText")}}
+              {{yield to="triggerText"}}
```
simonihmig, line 52: I don't have a strong opinion on string-based args vs. named blocks, but it seems this is a bit inconsistent now, as we we have `@label` (and no `{{yield to="label"}}`), but no `@triggerText`. 

Is there a reason for going this route here? Like do we expect trigger text to not just be a string, but some more complex HTML maybe?

Also, it doesn't seem to be mentioned in the docs! (only in the demo)

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,245 @@
...
+      name: 'sample.txt',
+      type: 'text/plain',
+    });
+
+    triggerEvent('[data-file-input-field]', 'change', { files: [file] });
+
+    await settled();
+
+    assert.verifySteps(['handleChange']);
+  });
```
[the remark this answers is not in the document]
simonihmig: Absolutely, love the file specific tests here!

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,274 @@
...
+    </template>);
+
+    const file = createFile(['Upload file sample'], {
+      name: 'sample.txt',
+      type: 'text/plain',
+    });
+
+    triggerEvent('[data-file-input-field]', 'change', { files: [file] });
+
+    await settled();
```
simonihmig, line 171: I think this could be simplified to

```suggestion
    await triggerEvent('[data-file-input-field]', 'change', { files: [file] });
```

Or didn't this work?

simonihmig (commented):
I left a bunch of comment, some things we might have to think about.

But this shouldn't deflect from the fact that this is _really_ some very nice work here! 😍

ember-toucan-core/src/components/form/file-input-field.hbs
nicolechung: Screen Shot 2023-03-09 at 12 35 42 PM

These are all good points! Perhaps I should revisit what I wrote in the docs.

For context, this is a bit easier to explain with pictures...the "Select files" is particular to file inputs, which have both a `label`, then a `button label` then a `description for the button` (like `Select files`)

So looks like
1. I should possibly rename `triggerText`, but not sure what
2. update docs to explain difference between `@label` and what ends up being `triggerText`

simonihmig: Thanks for that helpful explanation!

But just to be clear here, my point was not that we don't need both "triggerText" and label. My point was: is there a reason that triggerText is a named block (which could accept more than a plain string, like arbitrary HTML), but label is an argument (only accepting strings)?

At first it seems a bit inconsistent to me, but maybe there is a good reason for this?

ember-toucan-core/src/components/form/file-input-field.hbs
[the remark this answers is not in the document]
nicolechung: Update: `@onChange` and `@files` should be added here now along with a bunch of notes.

docs/components/file-input-field/index.md
```diff
@@ -0,0 +1,113 @@
...
+1. the list of files (as a FileList) from the target
+2. the raw event object
+
+It's most common to use this in combination with `@files` which will set the files (FileList) for the input based on the input received from the change event.
+
+```hbs
+<Form::FileInputField
+  @label='Label'
+  @onChange={{this.handleChange}}
+/>
```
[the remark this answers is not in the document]
nicolechung: - [x] added @files section
- [x] updated docs - onChange no longer returns anything, it returns `void` now.

docs/components/file-input-field/index.md
```diff
@@ -0,0 +1,106 @@
...
+import { tracked } from '@glimmer/tracking';
+
+export default class extends Component {
+  // Note that a File object is different from a FileList!
+  @tracked files: File[] = [];
+
+  @action
+  handleChange(e) {
+    console.log({ files: e.target.files });
+    // convert to a File[] by using the spread operator
```
nicolechung, line 51: Added comment here about using `...` (spread) operator.

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+            {{/if}}
+          </div>
+        </div>
+      </field.Control>
+    </field.Label>
+    {{#if @error}}
+      <field.Error id={{field.errorId}} data-error>{{@error}}</field.Error>
+    {{/if}}
+    <ul class="px-0" data-files>
```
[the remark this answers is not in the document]
nicolechung: Added a condition, and used `list-none` instead of `px-0`.

ember-toucan-core/src/components/form/file-input-field.hbs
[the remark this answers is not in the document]
nicolechung: And `@onDelete`

ember-toucan-core/src/components/form/file-input-field.ts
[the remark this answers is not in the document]
nicolechung: turns out `new FileList()` is not even allowed...to get around this, I have an internal `onChange` handler that extracts the files from the `FileList` in the target and passes that `File[]` to the `@onChange`.

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,112 @@
...
+      {{#each this.files as |file|}}
+        <li class="flex items-center p-0">
+          <section class="flex-1 p-0">
```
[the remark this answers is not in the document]
nicolechung: Think I switched it to a `div`...not sure why I used `section` I think it was an attempt to make it more semantic 😅

test-app/tests/integration/components/file-input-field-test.gts
nicolechung: I didn't know `triggerEvent` was async...TIL.

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,110 @@
...
+  <Form::Field as |field|>
+    <field.Label for={{field.id}} data-label>
+      {{@label}}
+      {{#if @hint}}
+        <field.Hint data-hint>{{@hint}}</field.Hint>
+      {{/if}}
+
+      <field.Control class="flex">
+        <div
+          class="bg-overlay-1 shadow-focusable-outline focus-within:shadow-focus-outline relative inline-flex h-8 w-full flex-wrap items-center rounded-sm"
```
ynotdraw, line 14: I'm wondering if we can move some of these classes from the `<div>` to `<field.Control` to prevent another wrapping div element?

Also wondering if some of these classes are needed at all?  For example, do we need `w-full` and `flex-wrap` here?  Our other inputs/textareas do _not_ set `w-full`, so you may want some padding.

```html
class="bg-overlay-1 shadow-focusable-outline focus-within:shadow-focus-outline relative inline-flex h-8 items-center rounded-sm"
```

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,110 @@
...
+            type="file"
+            accept={{this.accept}}
+            multiple={{this.multiple}}
+            {{! template-lint-disable no-bare-strings no-inline-styles  }}
+            style="clip: rect(1px, 1px, 1px, 1px);"
+            ...attributes
+            {{on "change" this.onChange}}
+          />
+          <div
+            class="text-titles-and-attributes inline-flex w-full cursor-pointer items-center gap-x-2 pl-2"
```
ynotdraw, line 35: I'm curious about the `pl-2` here.  Is that to offset the icon from the left-side of the box?  If so, should we remove this class and instead apply padding to the wrapping div (line 12/13) above instead?

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,326 @@
...
+        e.message.includes('A "@label" argument is required'),
+        'Expected assertion error message'
+      );
+    });
+
+    await render(<template>
+      {{! @glint-expect-error: we are not providing @label, so this is expected }}
+      <FileInputField @onChange={{onChange}} />
+    </template>);
+  });
```
ynotdraw, line 325: It looks like we require a few other arguments as well - should we write tests for these too?

```
trigger
onChange
onDelete
```

ember-toucan-core/src/components/form/file-input-field.hbs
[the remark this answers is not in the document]
ynotdraw: Ah! Up to you but I'd be happy to pair on this sometime this week!  Doesn't _have_ to be in this PR either so that this can be focused on the functionality and we can sweep through and make more nice-to-have downstream PRs later

docs/components/file-input-field/demo/base-demo.md
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function createFile(
+  content = ['Some sample content'],
+  options: Options = { name: '', type: '' }
+) {
+  const { name, type } = options;
+
+  const file = new File(content, name, {
+    type: type ? type : 'text/plain',
```
simonihmig, line 42: Nit: Could be simplified to 
```suggestion
    type: type ?? 'text/plain',
```

docs/components/file-input-field/index.md
simonihmig: I do not understand this, `files` is already of type `File[]`, isn't?

docs/components/file-input-field/index.md
```diff
@@ -0,0 +1,111 @@
...
+  @trigger='Select Files'
+  @rootTestSelector='example'
+/>
+```
+
+```js
+assert.dom('[data-root-field="example"]');
+
+// targeting this field's specific label
+assert.dom('[data-root-field="example"] > [data-label]');
```
simonihmig, line 84: This JS file is confusing me, is this representing a test (as the above line suggests), or the component class (as below)? Seems like a mix?

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,110 @@
...
+              >{{file.name}}</p>
+              <span
+                class="text-body-and-labels block"
+                data-file-size
+              >{{(this.formatSize file.size)}}</span>
+            </div>
+            <button
+              type="button"
+              class="flex-none bg-transparent"
+              {{on "click" (fn @onDelete file)}}
```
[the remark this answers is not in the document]
simonihmig: Both are great points!

test-app/tests/integration/components/file-input-field-test.gts
simonihmig: Yeah, pretty much all helpers from `@ember/test-helpers` are. You can think of them as having `await settled()` as their last statement internally, so you should never need another `settled()` call directly after those!

docs/components/file-input-field/demo/base-demo.md
```diff
@@ -0,0 +1,52 @@
+```hbs template
+<Form::FileInputField
+  @label='Label'
+  @files={{this.files}}
+  @hint='Hint text'
+  @trigger='Select Files'
+  @onChange={{this.handleChange}}
+  @onDelete={{this.handleDelete}}
```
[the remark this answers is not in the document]
simonihmig: Yeah, that aligns with my [earlier comment](https://github.com/CrowdStrike/ember-toucan-core/pull/64#discussion_r1130712912). We could have 
1. a single `@onChange`
2. `@onAdd` and `@onDelete`
3. `@onAdd`, `@onDelete` and `@onChange` (the latter being called for both add and delete events)

In most cases, as the user of our component is owning `@files`, they will want to simply update their tracked files array on `@onChange` (i.e. doesn't matter for them whether that came from adding or removing a file). So option 2. is a bit more inconvenient.

So if we think there are use cases where you would need to act differently based on whether the user action was adding or removing a file, then 3. would be handy. But can you think of such a use case? If not, option 1. is it I guess?

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  @action
+  onChange(event: FileEvent) {
+    if (event.target?.files) {
+      const files = [...event.target.files];
+
+      return this.args.onChange(files, event);
+    }
+
+    return this.args.onChange([], event);
```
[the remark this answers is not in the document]
nicolechung: I don't think it should ever happen, but I am pretty sure typescript/linter will complain...let me remove it and see if I can see the errror...

docs/components/file-input-field/ui-state-demo/base-demo.md
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  return file;
+}
+
+type Options = {
+  name: string,
+  type?: string,
+};
+```
```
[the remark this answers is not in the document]
nicolechung: So I was struggling a bit, because now `@onChange` and `@onDelete` are needed, which requires a backing class. So that's why I made the additional demo.

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,110 @@
...
+                  clip-rule="evenodd"
+                  d="M4.14648 11.5V5.5H5.14648V11.5H4.14648ZM7.14648 11.5V5.5H8.14648V11.5H7.14648Z"
+                  fill="#FAFAFA"
+                />
+                <path
+                  fill-rule="evenodd"
+                  clip-rule="evenodd"
+                  d="M6.14648 1C5.31806 1 4.64648 1.67157 4.64648 2.5H3.64648C3.64648 1.11929 4.76577 0 6.14648 0C7.5272 0 8.64648 1.11929 8.64648 2.5H7.64648C7.64648 1.67157 6.97491 1 6.14648 1Z"
+                  fill="#FAFAFA"
+                />
```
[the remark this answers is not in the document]
nicolechung: Screenshot 2023-03-15 at 10 58 34 AM

I updated the icon but this might have to do with the docfy overrides, maybe we can pair on this?

ember-toucan-core/src/components/form/file-input-field.hbs
nicolechung: Changed styles with @ynotdraw

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,110 @@
...
+                data-file-name
+              >{{file.name}}</p>
+              <span
+                class="text-body-and-labels block"
+                data-file-size
+              >{{(this.formatSize file.size)}}</span>
+            </div>
+            <button
+              type="button"
+              class="flex-none bg-transparent"
```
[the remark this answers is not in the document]
nicolechung: https://github.com/CrowdStrike/ember-oss-docs/issues/17

Adding in `reset-style` for now

test-app/tests/integration/components/file-input-field-test.gts
nicolechung: Should be in now.

docs/components/file-input-field/demo/base-demo.md
```diff
@@ -0,0 +1,52 @@
...
+    type: type ? type : 'text/plain',
+  });
+
+  return file;
+}
+
+type Options = {
+  name: string,
+  type?: string,
+};
```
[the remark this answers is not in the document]
nicolechung: I just tried switching to `ts` and it broke the build (guess we don't support that). Will remove.

ember-toucan-core/src/components/form/file-input-field.hbs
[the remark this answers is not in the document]
nicolechung: `aria-label` should be there now.

test-app/tests/integration/components/form/file-input/delete-button-test.gts
```diff
@@ -0,0 +1,65 @@
...
+    type: type ? type : 'text/plain',
+  });
+
+  return file;
+}
+
+module('Integration | Component | Form | FileInput | DeleteButton', function (hooks) {
+  setupRenderingTest(hooks);
+
+
```
simonihmig, line 35: Just wondering here: @nicolechung do you have prettier working for locally with gts files? It seems it _should_ have removed those double blank lines usually!

Also wondering, if that is not correctly formatted, then why is this not failing in CI? We do run `format:check:` there. Maybe something wrong with our prettier setup? 

Nothing to fix this PR IMO, but maybe we should look into this later.

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,39 @@
+<div
+  class="flex flex-col {{if @isDisabled 'text-disabled'}}"
+  data-root-field={{if @rootTestSelector @rootTestSelector}}
+>
+  <Form::Field as |field|>
+    <field.Label for={{field.id}}>
```
simonihmig, line 6: I just noticed by playing with this in https://da42f841.ember-toucan-core.pages.dev/docs/components/file-input-field, that having the label wrap the whole control and hint, it causes the whole block with a with of 100% to be clickable. For example, if you click _right of_ the upload button (i.e. not hitting the button!), then the file dialog will pop up nevertheless! 

I don't think that's what we want, right?

What was the reason for wrapping the label, instead of the usual approach we have elsewhere?

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,39 @@
...
+
+      <field.Control class="mt-1.5">
+        <Form::Controls::FileInput
+          aria-describedby="{{if this.hasError field.errorId}}"
+          aria-invalid={{if this.hasError "true"}}
+          @hasError={{this.hasError}}
+          accept={{this.accept}}
+          multiple={{this.multiple}}
+          id={{field.id}}
+          @onChange={{@onChange}}
```
simonihmig, line 20: Shouldn't this be `@onChange={{this.onChange}}`? Otherwise `this.onChange` would be unused.

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,94 @@
...
+    return Boolean(this.args?.error);
+  }
+
+  @action
+  onChange(event: FileEvent) {
+    if (!event.target?.files) {
+      return;
+    }
+
+    const files = [...event.target.files];
```
simonihmig, line 70: Something is wrong here (beyond `@onChange={{@onChange}}`, see my other comment). In the docs app I tried to select multiple files, but once I select a new file, it _replaces_ the existing one.

I think this should be something like this probably, to merge the new file(s) with the existing ones?

```suggestion
    const files = [...this.args.files, ...event.target.files];
```

But only when `this.args.multiple === true`, otherwise we indeed need to replace.

And we should add a test for multiple selection, to cover this!

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -0,0 +1,39 @@
...
+    {{#if @error}}
+      <field.Error id={{field.errorId}} data-error>{{@error}}</field.Error>
+    {{/if}}
+    {{#if @files.length}}
+      <Form::FileInput::List
+        @files={{@files}}
+        @onDelete={{this.handleDelete}}
+        @deleteLabel={{@deleteLabel}}
+      />
+    {{/if}}
```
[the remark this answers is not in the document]
nicolechung: > Here we don't don't have a yielded block though. Is that something that's still to come?

Yeah, so at the earlier point, I had already split the components. So I thought the easiest would be to leave the `split` but not yield any `blocks` or contextual components yet, but leave the smaller components in place for a follow up (if we decide to expose that).

test-app/tests/integration/components/form/file-input/delete-button-test.gts
nicolechung: Been running `pnpm format` over the packages manually, with the exception of the `markdown` files which I format in VS Code. I might have missed this, will run it again.

ember-toucan-core/src/components/form/controls/file-input.ts
```diff
@@ -0,0 +1,45 @@
...
+  onChange(event: FileEvent) {
+    if (!event.target?.files) {
+      return;
+    }
+
+    const files = [...event.target.files];
+
+    if (this.args.onChange) {
+      return this.args.onChange(files, event);
+    }
```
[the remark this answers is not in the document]
nicolechung: Should be updated now

ember-toucan-core/src/components/form/file-input/list.hbs
```diff
@@ -0,0 +1,22 @@
...
+    {{#if (has-block)}}
+      {{yield
+        (hash
+          ListItem=(component
+            (ensure-safe-component this.ListItem)
+            file=file
+            onDelete=@onDelete
+            deleteLabel=@deleteLabel
+          )
+        )
```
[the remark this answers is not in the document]
nicolechung: This is a good idea! `file` should be yielded now

ember-toucan-core/src/components/form/file-input-field.hbs
nicolechung: Okay this was a mistake on my part, but a bit messy so I am going to break down what I did (for my benefit, mostly)

1. in the controls/file-input, there is an `on` handler which takes a `this.onChange` handler. That handler only takes an `event` as a parameter, because it's using `on`.
2. So in `controls/file-input`, in the handler, I then call the `onChange` with the files array, AND the event. 
3. in `file-input-field` this should actually YES, be `@onChange={{@onChange}}`. What was wrong is I forgot to delete the `this.onChange` from the backing class.

ember-toucan-core/src/components/form/file-input-field.ts
nicolechung: Thank you for catching this!

1. new test for `multiple` has been added. Covers `adding` (`multiple`) and `replacing` (no `multiple` arg).
2. this whole `this.onChange` has been deleted from `file-input-field` (see explanation above...it wasn't in use)
3. the logic you suggested is now in `controls/file-input`.

test-app/tests/integration/components/form/file-input/delete-button-test.gts
simonihmig: Uh, you are right. Tried that in headless, it's the same there (the command is different, like `pnpm lint:prettier:fix`), prettier doesn't even run on .gts files. It's only VS Code that is doing this for me ootb...

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -274,6 +280,76 @@ module('Integration | Component | FileInputField', function (hooks) {
...
+      <FileInputField
+        @deleteLabel='Delete File'
+        @label="Label"
+        @trigger="Select Files"
+        @onChange={{realOnChange}}
+        @files={{ctx.currentFiles}}
+        @multiple={{true}}
+        data-file-input-field
+      />
+    </template>);
```
simonihmig, line 339: I would prefer to split this up into two separate test cases! (for when `@multiple` is false or true)

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,415 @@
...
+    
+    await triggerEvent('[data-file-input-field]', 'change', { files: [avocado, banana] });
+    
+    assert.dom('ul').exists('List of files exist');
+    assert.dom('li').exists('First list item exists');
+    assert.dom('ul > li:nth-child(2)').exists('there is more than 1 list element');
```
simonihmig, line 345:
```suggestion
    assert.dom('li').exists({ count: 2 }, 'Two list items exist');
```

ember-toucan-core/src/components/form/file-input-field.hbs
simonihmig: Hm, not sure about this. I still feel this should get solved in some way, but this can be done in a follup-up PR I think. This one already has so many comments! 😅 

Maybe we can find examples in the wild for how others have solved this?

nicolechung: I have an idea about this, will handle in a followup: https://github.com/CrowdStrike/ember-toucan-core/issues/86

ember-toucan-core/src/components/form/file-input-field.hbs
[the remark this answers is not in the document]
nicolechung: Moved this to: https://github.com/CrowdStrike/ember-toucan-core/issues/86

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -0,0 +1,429 @@
...
+    const file = createFile(['Upload file sample'], {
+      name: 'sample.txt',
+      type: 'text/plain',
+    });
+
+    await triggerEvent('[data-file-input-field]', 'change', { files: [file] });
+
+    await triggerEvent('button', 'click');
+
+    assert.dom('ul').doesNotExist();
```
ynotdraw, line 290: Nit: To ensure the file is added and then removed and to leverage some of the existing data attributes I think something like this would work:

```suggestion

    // Verify no files
    assert.dom('[data-files]').doesNotExist();
    
    await triggerEvent('[data-file-input-field]', 'change', { files: [file] });
    
    // Verify files are there
    assert.dom('[data-files]').exists();

    await triggerEvent('button', 'click');

    // Verify the `ul` is gone as all files are deleted
    assert.dom('[data-files]').doesNotExist();
```

test-app/tests/integration/components/file-input-field-test.gts
ynotdraw, line 1: Since `gts/gjs` doesn't get covered with prettier fix yet, you may want to open all of these test files up in VSCode and save them so that prettier auto-formats them.  Otherwise the next time someone comes into one of these tests who does use VSCode hits save, there will be a bunch of prettier fixes added.

ynotdraw (commented):
This is looking good to me!  Maybe only the "major" thing that should likely be addressed before merging is the `package.json` change that can be resolved by running `pnpm build` and then checking in those changes as mentioned in the comment.

I'd like to cut a release in this repo in the next day or two, so depending on how you're feeling I think we have a few options.  I'd like to get this merged!

1. Merge this in to `main`, continue iterating, knowing that this component still has a few updates to go and may require a few releases quite soon to iron out some issues.
2. Create a new feature branch off of `main`, update this PR to target that branch instead, so then you can merge into the feature branch instead of `main` directly.  Then any subsequent changes will also target the feature branch.  Whenever you're feeling like everything is covered, then put up the feature branch for PR to merge into `main` - since we already reviewed all feature branch PRs, this should be a stamp of approval PR.  That way this won't get added into `main` until you're 100% ready.

I'm good with either option since we are still pre1.0.0.  Let me know what you think! Nice work!

test-app/tests/integration/components/file-input-field-test.gts
nicolechung: Done in this commit: [2f677f7](https://github.com/CrowdStrike/ember-toucan-core/pull/64/commits/2f677f73032868ea1f6f029a558f599ba990d155)

ember-toucan-core/src/components/form/controls/file-input.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface ToucanFormControlsFileInputComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    hasError?: boolean;
+    trigger: string;
+    isDisabled?: boolean;
+    onChange?: (files: File[] | [], event: FileEvent) => void;
+    multiple?: boolean;
+    files?: File[];
```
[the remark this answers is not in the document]
nicolechung: going to do this as a follow up!

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  get accept() {
+    return this.args.accept ?? '*';
+  }
+
+  get multiple() {
+    return this.args.multiple ?? true;
+  }
+
```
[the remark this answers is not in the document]
nicolechung: Filed under: https://github.com/CrowdStrike/ember-toucan-core/issues/91

noahgelmancs (approved):

PR #65, opened 2023-03-06
[Don't Merge!] Design error+ focus testing

ynotdraw:
# DO NOT MERGE! For testing only!

PR #63, opened 2023-03-06
Implement design suggested spacing of elements

ynotdraw:
## 💅  Description
This implements the styling changes after a design review.

---

## 🔬 How to Test

- View https://35eff113.ember-toucan-core.pages.dev/docs/components/textarea-field
- The hint block + control + error block should all have 0.125rem/2px spacing _above_ the elements
- View https://35eff113.ember-toucan-core.pages.dev/docs/components/checkbox-field
- The hint block + error block should have 0.125rem/2px spacing _above_ the elements
- View https://7bfeb461.ember-toucan-core.pages.dev/docs/components/input-field

---

## 📸 Images/Videos of Functionality

Screenshot 2023-03-06 at 9 00 45 AM

PR #66, opened 2023-03-06
fix: Update input-field to use hasError and apply the error shadow

ynotdraw:
## 🐛  Description
Fix a styling issue with `input field` that I noticed when doing https://github.com/CrowdStrike/ember-toucan-core/pull/65/files.

Downstream of https://github.com/CrowdStrike/ember-toucan-core/pull/63

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-03-06 at 3 00 37 PM  | Screenshot 2023-03-06 at 2 56 37 PM  |
| No error border  | With error border  |

---

## 🔬 How to Test

- View https://e963e8f6.ember-toucan-core.pages.dev/docs/components/input-field#inputfield-with-label-and-error
- Verify error shadow on errored input

### DIFF
--- a/test-app/tests/integration/components/input-field-test.gts
+++ b/test-app/tests/integration/components/input-field-test.gts
@@ -21,7 +21,7 @@ module('Integration | Component | InputField', function (hooks) {
 
     const label = '[data-label]'
     const input = '[data-input]';
-    const inputId = find(input)?.getAttribute('id') || ''; 
+    const inputId = find(input)?.getAttribute('id') || '';
 
     assert.ok(inputId, 'Expected to have id');
     assert.dom(label).exists('Expected to have label block rendered');
@@ -30,6 +30,7 @@ module('Integration | Component | InputField', function (hooks) {
     assert.dom(input).exists('Expected to have input tag rendered');
     assert.dom(input).hasAttribute('type', 'text');
     assert.dom(input).hasAttribute('id');
+    assert.dom(input).hasNoClass('shadow-error-outline');
   });
 
   test('it encourages @label via assert', async function (assert) {
@@ -39,14 +40,14 @@ module('Integration | Component | InputField', function (hooks) {
       assert.strictEqual(e.message, 'Assertion Failed: input field requires a label', 'Expected an error message if @label is not provided')
     })
     await render(<template>
-      {{! @glint-expect-error: should have an error here for missing @label }} 
+      {{! @glint-expect-error: should have an error here for missing @label }}
       <InputField
         type="text"
         />
     </template>);
 
   })
-  
+
   test('it renders an error', async function (assert) {
     await render(<template>
       <InputField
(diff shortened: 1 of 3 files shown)


test-app/tests/integration/components/input-field-test.gts
```diff
@@ -21,7 +21,7 @@ module('Integration | Component | InputField', function (hooks) {
 
     const label = '[data-label]'
     const input = '[data-input]';
-    const inputId = find(input)?.getAttribute('id') || '';
```
ynotdraw, line 24: Prettier picked up on a bunch of changes, but I removed them and only applied the empty spacing changes here

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -30,6 +30,7 @@ module('Integration | Component | InputField', function (hooks) {
     assert.dom(input).exists('Expected to have input tag rendered');
     assert.dom(input).hasAttribute('type', 'text');
     assert.dom(input).hasAttribute('id');
+    assert.dom(input).hasNoClass('shadow-error-outline');
```
nicolechung, line 33: ✅

test-app/tests/integration/components/input-field-test.gts
nicolechung: Thanks for this! Um...how did you selectively apply the prettier changes??

nicolechung (approved):

test-app/tests/integration/components/input-field-test.gts
simonihmig: Hm, so `prettier --check` doesn't run in CI then it seems?

simonihmig (approved):

test-app/tests/integration/components/input-field-test.gts
ynotdraw: This is super interesting.  So `prettier --check` is returning as all is good.  Running `pnpm format`, I see the following:

```bash
➜  test-app git:(fix-input-error-outline) ✗ pnpm format

> test-app@0.0.0 format /workspace/ember-toucan-core/test-app
> prettier -w .

.eslintrc.cjs 23ms
.prettierrc.js 5ms
.template-lintrc.js 3ms
.watchmanconfig 1ms
app/app.ts 79ms
app/config/environment.d.ts 6ms
app/index.html 16ms
app/router.ts 3ms
app/styles/app.css 7ms
app/templates/application.hbs 17ms
config/ember-cli-update.json 2ms
config/ember-try.js 5ms
config/environment.js 4ms
config/optional-features.json 1ms
config/targets.js 2ms
ember-cli-build.js 2ms
package.json 2ms
README.md 24ms
testem.js 3ms
tests/helpers/index.ts 4ms
tests/index.html 4ms
tests/test-helper.ts 2ms
tsconfig.json 1ms
types/global.d.ts 2ms
types/test-app/index.d.ts 2ms
```

Which is interesting, as prettier seems to be skipping over `.gjs/gts` files.  So it seems like in VSCode, prettier auto formats things as you'd expect, but these files are getting missed via the CLI.  Any ideas?

Nothing really sticks out to me in the [prettierrc file](https://github.com/CrowdStrike/ember-toucan-core/blob/main/test-app/.prettierrc.js).  We have an override for `.gjs/gts` files, but I can't seem to get any of [these rules](https://prettier.io/docs/en/options.html) to run against this extension.

PR #69, opened 2023-03-07
#67: Add radio + radiogroupfield component

### LINKED ISSUE #67: Add RadioGroupField component
We should have a RadioGroupField component that handles the radio group use case.

ynotdraw:
## 🚀 Description
This PR introduces a `Radio` control component, a `RadioField` component, and a `RadioGroupField` component.   It follows the same pattern as `Checkbox` + `CheckboxField`: https://github.com/CrowdStrike/ember-toucan-core/pull/50

Closes #67 

---

## 🔬 How to Test

- View https://25c29909.ember-toucan-core.pages.dev/docs/components/radio
- View https://25c29909.ember-toucan-core.pages.dev/docs/components/radio-field
- View https://cd7b4e7b.ember-toucan-core.pages.dev/docs/components/radio-group-field
  - **NOTE:** These components are not 100% styled yet, as we are still figuring out where to land on a few things.  This is about ~90% close though.

---

Screenshot 2023-03-07 at 2 29 06 PM

### a11y test

**NOTE:** It's saying 14 radio buttons as I am rendering all available UI states below so it's counting all of those 🙈 

https://user-images.githubusercontent.com/8069555/223532600-bb0e2458-326a-4a59-84df-def5aba411ad.mov


ember-toucan-core/src/components/form/radio-field.hbs
```diff
@@ -0,0 +1,42 @@
...
+      </label>
+    </field.Control>
+
+    {{#if @error}}
+      <field.Error
+        class="mt-0.5"
+        id={{field.errorId}}
+        data-error
+      >{{@error}}</field.Error>
+    {{/if}}
```
ynotdraw, line 40: I don't think anyone will actually use this individual radio error over what we're planning to provide via the `Fieldset`-wrapped option, but I _suppose_ they could.  We're still ironing things out from a design perspective here, so we may remove this after further design review, or it may stay!  Here's what the Fieldset option may look like:

```hbs
<Form::Fieldset
  @label='Label'
  @hint='Extra information about the fieldset'
  @error='Error message'
>
    <Form::RadioField @value="option1" @label='Label1' name="label1" />
    <Form::RadioField @value="option2" @label='Label2' name="label2" />
    <Form::RadioField @value="option3" @label='Label3' name="label3" />
</Form::Fieldset>
```

And then we get the error border/shadow for free as seen at: https://25c29909.ember-toucan-core.pages.dev/docs/components/fieldset#all-ui-states

simonihmig (commented):
Code-wise this looks good! _But_...

I have a bit of a problem or question wrt to what we define as the "field" here. At least the radio field does not map to what a radio field is in headless-form. Which does not mean the one is "right" and the other is "wrong", but at least we should see if/how we can align this maybe and fill the gap...

Take [this example](https://88240223.ember-headless-form.pages.dev/docs/validation/native), where we have a form with three fields (in the sense that I have been using that term): name, email and gender. So "field" is basically an element of the form that the user can assign (enter or select) a value to. And it has a `@name` that will map to the form data object. What is inherently special about a radio group field here is that a single selection (here "gender") is represented by _multiple_ radio inputs, unlike all the other controls where this is a 1:1 mapping. You could represent the gender selection as well using a _single_ `<select>` control. But in this sense I would see the _group_ of _multiple_ radio buttons as the "field". Having a selection with a single radio button makes no sense after all. 

This is different to what we have here, as here a "radio field" represents a single radio button (and related label and stuff).

So one thing is the naming here, which we probably can change easily. But also I wonder if we need something that represents that "radio group" as a whole (the PR title even mentions "radiogroupfield", while I don't see that here, right?). So something that matches what `field.Radio` does in headless-form. Off the top of my head, that one could:
* yield a contextual `.RadioField` component (the one we already have here)
* accept a single `@onChange` and curry that to the yielded `.RadioField`
* accept a `@value` which is _not_ the value of the single radio button, but basically the user-selected value. This would match what e.g. an `InputField` would have, as that also has `@value` representing the entered value. We don't have an equivalent to that here yet AFAICT, right?
* apply the `eq` logic ootb, as it has the selected `@value` as well as the value of each individual radio button (yielded RadioField)
* apply the same `name` to each radio button.

Regarding the last point of `name`, this opens another can of worms. So first, assigning a `name` to each form control is not only best-practice for semantic form markup (as that is how forms work without JS), it is also a requirement (or strong recommendation at least) for using this with headless-form (you probably remember those discussions around triggering validations and `captureEvent` in case the control has no `name`). 

For radio groups that is especially important, as the common name basically defines what the group is, so you get the mutually exclusive behaviour of a radio group. With our JS-based solution, we can actually get around that, as we do those `eq` checks in JS, but still I would consider it strongly recommended to have a name applied. I believe that could also actually fix the screen reader problem you were seeing, as when the radio buttons have no name they share, then there is no way for the user agent to determine what the group is they belong to!

So, with a plain input, we can just pass the `name` as an attribute as in `<InputField name="foo"/>`, but this wouldn't work for a radio group as in `<RadioGroup name="foo">`, because we would need to spread that name to all individual radio inputs in that group, and you cannot do that with HTML attributes (there is only a singe `...attributes`). headless-form uses an `@name` argument, which it can ["curry" to all radio inputs](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/control/radio.hbs#L7). Maybe we have to introduce that argument here also? (and for consistency across all other fields)

test-app/tests/integration/components/radio-test.gts
```diff
@@ -0,0 +1,156 @@
...
+      {{! template-lint-disable require-input-label }}
+      <RadioControl @value="option" @isDisabled={{true}} data-radio />
+    </template>);
+
+    assert.dom('[data-radio]').isDisabled();
+
+    assert.dom('[data-radio]').hasClass('border-disabled');
+    assert.dom('[data-radio]').hasClass('bg-transparent');
+    assert.dom('[data-radio]').doesNotHaveClass('bg-primary-idle');
+    assert.dom('[data-radio]').doesNotHaveClass('border-none');
```
nicolechung, line 42: Suggestion: do you have to also had a check for not having `bg-disabled`?
https://github.com/CrowdStrike/ember-toucan-core/pull/69/files#diff-d84c4bf7f661538e00fe281442c5504a11ae3dc2ca5d55aa7ed07b3067972746R32

test-app/tests/integration/components/radio-test.gts
```diff
@@ -0,0 +1,156 @@
...
+  });
+
+  test('it makes the radio checked using `@isChecked={{true}}`', async function (assert) {
+    await render(<template>
+      {{! we do not require a label, but instead suggest using Field }}
+      {{! template-lint-disable require-input-label }}
+      <RadioControl @value="option" @isChecked={{true}} data-radio />
+    </template>);
+
+    assert.dom('[data-radio]').isChecked();
```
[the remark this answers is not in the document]
nicolechung: Update: I see it's over here: https://github.com/CrowdStrike/ember-toucan-core/pull/69/files#diff-c236e8ab9d8ea1ccfed5a6f92cdac501a0724541c36966360a78c82b756b6dcdR63

Maybe remove the class checks in the first two tests and leave it in the tests for `expected classes`?

ynotdraw:
@simonihmig Thanks a bunch for your thoughts! I definitely agree here.  I was thinking at first that this could be the beginnings of the `RadioGroup/RadioGroupField` component, but then decided to punt on it until we get more forms stuff wired up.  However, after this discussion, I 100% agree that we should handle this here and now, as part of this PR as it'll make our lives easier to solve this now.

Based on above, your comments, and thinking out loud maybe we can do this:

- Keep `RadioField` as it is now.  It essentially combines `Field` + `Radio` components
- Add `RadioGroup`/`RadioGroupField` (?) which will handle all of the logic in your bullet points above

So:

> - yield a contextual `RadioField` component (the one we already have here)
> - accept a single `@onChange` and curry that to the yielded `.RadioField`
> - accept a `@value` which is not the value of the single radio button, but basically the user-selected value. This would match what e.g. an InputField would have, as that also has `@value` representing the entered value. We don't have an equivalent to that here yet AFAICT, right?
> - apply the `eq` logic ootb, as it has the selected `@value` as well as the value of each individual radio button (yielded `RadioField`)
> - apply the same name to each radio button


Some rough pseudocode maybe:

```hbs
{{!-- RadioGroup / RadioGroupField --}}

<Form::Fieldset
  @label={{@label}}
  @hint={{@hint}}
  @error={{@error}}
>
    {{yield (hash RadioField=(ensure-safe-component this.RadioField) name=@name onChange=this.handleChange isSelected={{this.checkRadioGroupValueVersusThisRadiosValue)}}
</Form::Fieldset>
```

And then usage of above would be something like:

```hbs
<RadioGroup @label="Label" @name="options" @value="option-1" @onChange={{this.radioGroupChange}} as |group|>
  <group.RadioField @label="Opt 1" @value="option-1" />
  <group.RadioField @label="Opt 2" @value="option-2" />
  <group.RadioField @label="Opt 3" @value="option-3" />
</RadioGroup>
```

What do you think?

---

Whatever we decide here, I'll also apply to the checkbox group.  It's a little different but has similar issues!

test-app/tests/integration/components/radio-test.gts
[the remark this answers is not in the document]
ynotdraw: Okay I think I got this one taken care of!

test-app/tests/integration/components/radio-test.gts
ynotdraw: Thank ya! Got this updated too!

ynotdraw:
Looks like the cloudflare action is failing because of https://github.com/cloudflare/pages-action/issues/66 👀

docs-app/package.json
```diff
@@ -38,10 +38,10 @@
...
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
-    "@glint/core": "^0.9.6",
-    "@glint/environment-ember-loose": "^0.9.6",
-    "@glint/environment-ember-template-imports": "^0.9.6",
-    "@glint/template": "^0.9.6",
+    "@glint/core": "^1.0.0-beta.3",
+    "@glint/environment-ember-loose": "^1.0.0-beta.3",
+    "@glint/environment-ember-template-imports": "^1.0.0-beta.3",
+    "@glint/template": "^1.0.0-beta.3",
```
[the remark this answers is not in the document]
nicolechung: How did you figure this out that it needed updating?!

ynotdraw:
Thanks @NullVoxPopuli ! 🙇

docs-app/package.json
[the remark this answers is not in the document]
ynotdraw: Stumbled around for a bit longer than I care to admit ( 🙈 ), BUT!

```
test-app lint: [lint:types]   ../node_modules/.pnpm/@glint+template@0.9.7_@glimmer+component@1.1.2/node_modules/@glint/template/-private/integration.d.ts:21:64
```

I noticed `0.9.7` in the error log from the link above.  So I did a search of `@glint/template` and noticed `docs-app` was the only one not on beta-3.  Updated it and it all worked!

simonihmig:
> Maybe I should update to cloudflare/pages-action@v1.4.0?

> I just did that an hour or so ago on ember-resources, can confirm it works. 🎉

Yeah, probably works, but sound like a bug to me, in their "publishing" process. AFAIK `someGitHubAction@1` basically means `^1.0.0`, so you'll get non-breaking updates automatically. `cloudflare/pages-action@v1.4.0` will pin it to _exactly_ 1.4.0., so you'll never get newer releases (unless you or renovate updates the version specifier)

ynotdraw:
Okay this should be all good to review now!

ember-toucan-core/src/components/form/radio-field.ts
```diff
@@ -0,0 +1,40 @@
...
+    onChange?: ToucanFormRadioControlComponentSignature['Args']['onChange'];
+    /**
+     * A test selector for targeting the root element of the field. In this case, the wrapping div element.
+     */
+    rootTestSelector?: string;
+    /**
+     * This component argument is used to determine if the underlying radio is checked.
+     * When `selectedValue` and `value` are equal, the radio will have the checked attribute applied.
+     */
+    selectedValue?: string;
```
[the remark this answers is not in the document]
simonihmig: It would be pretty much pointless to try to disagree here, given that this is _exactly_ what I am doing in headless [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/control/radio.ts#L25) and [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/control/radio.ts#L56-L58) 😅

ember-toucan-core/src/components/form/radio-group-field.hbs
```diff
@@ -0,0 +1,19 @@
...
+  ...attributes
+>
+  {{yield
+    (hash
+      RadioField=(component
+        (ensure-safe-component this.RadioFieldComponent)
+        name=@name
+        onChange=this.handleInput
+        selectedValue=@value
+        isDisabled=@isDisabled
```
simonihmig, line 15: This is also nice to have by having a group, to disable all radio buttons at once! 🚀

ember-toucan-core/src/components/form/radio-group-field.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import RadioFieldComponent from './radio-field';
+
+import type { ToucanFormRadioFieldComponentSignature } from './radio-field';
+import type { WithBoundArgs } from '@glint/template';
+
+export interface ToucanFormRadioGroupFieldComponentSignature {
+  Element: HTMLFieldSetElement;
+  Args: {
+    error?: string;
```
simonihmig, line 12: Doesn't have to be in this PR, but just remembered @nicolechung mentioning this in one of her reviews of some headless PR: would be nice to have some JSDoc comments here explaining the args in a short sentence. As this is what people will see when having Glint enabled. And probably they will more likely look only at that, and not consult the actual docs, as we all know how docs get (not) read... 😅

Created: https://github.com/CrowdStrike/ember-toucan-core/issues/73

test-app/tests/integration/components/radio-group-field-test.gts
```diff
@@ -0,0 +1,137 @@
+/* eslint-disable no-undef -- Until https://github.com/ember-cli/eslint-plugin-ember/issues/1747 is resolved... */
+/* eslint-disable simple-import-sort/imports,padding-line-between-statements,decorator-position/decorator-position -- Can't fix these manually, without --fix working in .gts */
+
+import { click, render } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+
+import RadioGroupField from '@crowdstrike/ember-toucan-core/components/form/radio-group-field';
+import { setupRenderingTest } from 'test-app/tests/helpers';
+
+module('Integration | Component | RadioGroupField', function (hooks) {
```
simonihmig, line 10: Nice test coverage here!

simonihmig (approved):
Beautiful! 😍

One note to leave here though: it seems https://github.com/CrowdStrike/ember-headless-form/issues/22 also applies here to some extend, right? Like at least the point of where to apply `aria-invalid` is also not solved here AFAICT, right?

For `aria-describedby`, this is now handled by the fieldset component here, so `aria-describedby` is attached to the `<fieldset>` in case of errors, correct?

docs-app/package.json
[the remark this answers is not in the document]
nicolechung: good to know!

ember-toucan-core/src/components/form/radio-group-field.hbs
```diff
@@ -0,0 +1,19 @@
...
+      )
```
nicolechung, line 16: Suggestion: Shouldn't we also yield the `selectedValue` in case the user wants to format that in some way?

nicolechung (approved):

ember-toucan-core/src/components/form/radio-group-field.hbs
nicolechung: NVM - read the thread above!

ember-toucan-core/src/components/form/radio-field.ts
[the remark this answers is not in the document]
ynotdraw: I love it.  I'm going to do this as part of a follow up PR as this PR is blowing up quite a bit anyway.  Update coming soon!

ynotdraw:
> One note to leave here though: it seems https://github.com/CrowdStrike/ember-headless-form/issues/22 also applies here to some extend, right? Like at least the point of where to apply aria-invalid is also not solved here AFAICT, right?

Yup yup, that's correct.  Whatever @noahgelmancs finds there I'll apply here as well!

> For aria-describedby, this is now handled by the fieldset component here, so aria-describedby is attached to the <fieldset> in case of errors, correct?

Yup that's correct! Will need to verify that is okay as well!

PR #74, opened 2023-03-13
feat: Require RadioGroupField value component argument

ynotdraw:
## 🚀 Description

Make the `@value` argument required for `RadioGroupField`.

---

## 🔬 How to Test

- CI should pass

---

N/A

### DIFF
--- a/ember-toucan-core/src/components/form/radio-group-field.ts
+++ b/ember-toucan-core/src/components/form/radio-group-field.ts
@@ -15,11 +15,7 @@ export interface ToucanFormRadioGroupFieldComponentSignature {
     label: string;
     name: string;
     onChange?: ToucanFormRadioFieldComponentSignature['Args']['onChange'];
-    /**
-     * A test selector for targeting the root element of the field. In this case, the wrapping div element.
-     */
-    rootTestSelector?: string;
-    value?: ToucanFormRadioFieldComponentSignature['Args']['value'];
+    value: ToucanFormRadioFieldComponentSignature['Args']['value'];
   };
   Blocks: {
     default: [
--- a/ember-toucan-core/src/components/form/radio-group-field.ts
+++ b/ember-toucan-core/src/components/form/radio-group-field.ts
@@ -1,4 +1,5 @@
 import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 
 import RadioFieldComponent from './radio-field';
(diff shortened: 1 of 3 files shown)


ember-toucan-core/src/components/form/radio-group-field.ts
```diff
@@ -15,11 +15,7 @@ export interface ToucanFormRadioGroupFieldComponentSignature {
     label: string;
     name: string;
     onChange?: ToucanFormRadioFieldComponentSignature['Args']['onChange'];
-    /**
-     * A test selector for targeting the root element of the field. In this case, the wrapping div element.
-     */
-    rootTestSelector?: string;
-    value?: ToucanFormRadioFieldComponentSignature['Args']['value'];
+    value: ToucanFormRadioFieldComponentSignature['Args']['value'];
```
ynotdraw, line 18: I also tried adding an assert like:

```ts
  constructor(
    owner: unknown,
    args: ToucanFormRadioGroupFieldComponentSignature['Args']
  ) {
    assert('A "@value" argument is required', value);
    super(owner, args);
  }
```

but was getting errors in the docs-app on render/mount.  I'm thinking that due to the way yielded components work or something that `@value` isn't supplied or something on first render, but then shortly thereafter? I have 0 evidence this is the case but wanted to point out that I tried this and it didn't work 😭

ember-toucan-core/src/components/form/radio-group-field.ts
```diff
@@ -15,11 +15,7 @@ export interface ToucanFormRadioGroupFieldComponentSignature {
     label: string;
     name: string;
     onChange?: ToucanFormRadioFieldComponentSignature['Args']['onChange'];
-    /**
-     * A test selector for targeting the root element of the field. In this case, the wrapping div element.
-     */
-    rootTestSelector?: string;
```
ynotdraw, line 21: I removed this as it was copy pasta 🍝

ember-toucan-core/src/components/form/radio-group-field.ts
nicolechung: Did you try adding `this.args` before `value` in the assert?

Like this:
```
assert('A "@value" argument is required', this.args.value);
```

nicolechung (approved):

ember-toucan-core/src/components/form/radio-group-field.ts
ynotdraw: ... I am so dumb... I forgot to update the UI States section so _those_ were throwing errors and causing this to fail. Nothing to see here! 🙈

ynotdraw:
> Why are we doing this? We do not require an input to have a value (i.e. it can be empty), so why for radios? You can have a bunch of radios, all unchecked, see the example you gave here: https://adrianroselli.com/2022/02/support-for-marking-radio-buttons-required-invalid.html#R6

I guess similar to https://github.com/CrowdStrike/ember-toucan-core/pull/69#discussion_r1132115801, but up one level higher.  From our design guidance, we want folks to have an option selected by default.  Can share a link via Slack about the guidance.

ynotdraw:
Going to go ahead and close this after some discussion.  We don't want to enforce a selected value for radios!

PR #75, opened 2023-03-14
fix: Add aria-invalid + role="radiogroup" to RadioGroupField

ynotdraw:
## 🚀 Description

For accessibility and after reading through https://adrianroselli.com/2022/02/support-for-marking-radio-buttons-required-invalid.html#R6 again, it seems like this is valid and helps Voice Over announcing the error state.

---

## 🔬 How to Test

- Tests should pass!

---

N/A - no visual changes

### DIFF
--- a/ember-toucan-core/src/components/form/radio-group-field.hbs
+++ b/ember-toucan-core/src/components/form/radio-group-field.hbs
@@ -4,6 +4,7 @@
   @error={{@error}}
   @isDisabled={{@isDisabled}}
   aria-invalid={{if @error "true"}}
+  aria-required="true"
   role="radiogroup"
   ...attributes
 >
--- a/ember-toucan-core/src/components/form/radio-group-field.hbs
+++ b/ember-toucan-core/src/components/form/radio-group-field.hbs
@@ -3,6 +3,8 @@
   @hint={{@hint}}
   @error={{@error}}
   @isDisabled={{@isDisabled}}
+  aria-invalid={{if @error "true"}}
+  role="radiogroup"
   ...attributes
 >
   {{yield
--- a/test-app/tests/integration/components/radio-group-field-test.gts
+++ b/test-app/tests/integration/components/radio-group-field-test.gts
@@ -16,8 +16,17 @@ module('Integration | Component | RadioGroupField', function (hooks) {
     </template>);
 
     assert.dom('[data-group-field]').exists();
+    assert.dom('[data-group-field]').hasNoAttribute('aria-invalid');
 
     assert.dom('[data-label]').hasText('Label');
+  });
+
+  test('it sets "role" by default', async function (assert) {
+    await render(<template>
+      <RadioGroupField @label="Label" @name="group" data-group-field />
+    </template>);
+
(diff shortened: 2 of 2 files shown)


ember-toucan-core/src/components/form/radio-group-field.hbs
```diff
@@ -3,6 +3,8 @@
   @hint={{@hint}}
   @error={{@error}}
   @isDisabled={{@isDisabled}}
+  aria-invalid={{if @error "true"}}
+  role="radiogroup"
```
ynotdraw, line 7: Based on https://adrianroselli.com/2022/02/support-for-marking-radio-buttons-required-invalid.html#R6

> This is valid HTML and ARIA. Please bear in mind that the [radiogroup](https://www.w3.org/TR/wai-aria-1.1/#radiogroup) role is strictly for cases where only one item in a set can be selected at a time (so not checkboxes).

So since we are radios here, adding the `role` should be fine! We can't replicate this with grouped-checkboxes though.

nicolechung: Based on the example, doesn't `aria-required` need to be added in also?

test-app/tests/integration/components/radio-group-field-test.gts
```diff
@@ -134,4 +143,17 @@ module('Integration | Component | RadioGroupField', function (hooks) {
...
+    await render(<template>
+      <RadioGroupField
+        @label="Label"
+        @name="group"
+        @error="Error message"
+        data-group-field
+      />
+    </template>);
+
+    assert.dom('[data-group-field]').hasAttribute('aria-invalid', 'true');
```
nicolechung, line 157: Same question here: doesn't `aria-required` also need to be here?

ember-toucan-core/src/components/form/radio-group-field.hbs
ynotdraw: According to https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role#associated_wai-aria_roles_states_and_properties

> aria-required
> Indicates that one radio within the group must have aria-checked="true" set before the form may be submitted. The required state is specified on the radiogroup element rather than one of the radio elements, unlike when using HTML radio buttons where the [required](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) attribute is set directly on one or more radio [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) elements.


I'm not sure! We require a `value` as part of https://github.com/CrowdStrike/ember-toucan-core/pull/74, so they can't get in a state where a value isn't provided/automatically set.  Those docs explicitly mention `aria-checked` as well, where we use the `checked` attribute, so I'm not super sure it applies.  Normally `aria-checked` is used when _not_ using semantic HTML for radios/checkboxes, but here we are.  What do you think?

simonihmig (approved):
That seems to match what I did in https://github.com/CrowdStrike/ember-headless-form/pull/76!

nicolechung (approved):

PR #76, opened 2023-03-14
Design spacing iteration

ynotdraw:
## 🚀 Description
This adds suggested design spacing based on our guidelines.

---

## 🔬 How to Test

- The spacing between the label and the hint should be 2px
- The spacing between the hint and the control should be 6px
- The spacing between the control and the error should be 6px
- If there is no hint, the spacing between the label and the control should be 6px

---

Screenshot 2023-03-14 at 10 37 06 AM

### DIFF
--- a/ember-toucan-core/src/components/form/input-field.hbs
+++ b/ember-toucan-core/src/components/form/input-field.hbs
@@ -1,15 +1,19 @@
 <div
-  class="flex flex-col space-y-0.5 {{if @isDisabled 'text-disabled'}}"
+  class="flex flex-col {{if @isDisabled 'text-disabled'}}"
   data-root-field={{if @rootTestSelector @rootTestSelector null}}
 >
   <Form::Field as |field|>
     <field.Label for={{field.id}} data-label>{{@label}}</field.Label>
 
     {{#if @hint}}
-      <field.Hint id={{field.hintId}} data-hint>{{@hint}}</field.Hint>
+      <field.Hint
+        class="mt-0.5"
+        id={{field.hintId}}
+        data-hint
+      >{{@hint}}</field.Hint>
     {{/if}}
 
-    <field.Control class="flex">
+    <field.Control class="mt-1.5 flex">
       <Form::Controls::Input
         id={{field.id}}
         aria-describedby="{{if @hint field.hintId}} {{if @error field.errorId}}"
@@ -22,7 +26,11 @@
       />
     </field.Control>
     {{#if @error}}
-      <field.Error id={{field.errorId}} data-error>{{@error}}</field.Error>
+      <field.Error
+        class="mt-1.5"
+        id={{field.errorId}}
+        data-error
+      >{{@error}}</field.Error>
     {{/if}}
   </Form::Field>
 </div>
--- a/ember-toucan-core/src/components/form/input-field.hbs
+++ b/ember-toucan-core/src/components/form/input-field.hbs
@@ -6,11 +6,7 @@
     <field.Label for={{field.id}} data-label>{{@label}}</field.Label>
 
     {{#if @hint}}
-      <field.Hint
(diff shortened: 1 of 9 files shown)


ember-toucan-core/src/components/form/input-field.hbs
```diff
@@ -9,7 +9,7 @@
       <field.Hint id={{field.hintId}} data-hint>{{@hint}}</field.Hint>
     {{/if}}
 
-    <field.Control class="flex">
+    <field.Control class="mt-1.5 flex">
```
ynotdraw, line 12: It's tempting to add `mt-1.5` directly to `field.Control`, but we have cases where that'll actually cause unwanted padding like with Checkbox, RadioField, etc.

nicolechung:
I'm not seeing the double outline, am I looking at the right preview link?

Screen Shot 2023-03-14 at 2 42 55 PM

ynotdraw:
> I'm not seeing the double outline, am I looking at the right preview link?
> 
> Screen Shot 2023-03-14 at 2 42 55 PM

Good eye! This PR is only for the spacing between elements.  I'll have another PR up tomorrow to add the error+focus styles like you see in https://github.com/CrowdStrike/ember-toucan-core/pull/65/files#diff-8f38e0805927a539bc7a9b51bca3086e0a4fcfe1b075e051ab83500786594cb5R11.  I need to merge https://github.com/CrowdStrike/tailwind-toucan-base/pull/216, but I want to babysit it through the release process and I have to run in a few!

danwenzel (approved):
Code LGTM, but like Nicole, I couldn't find where to test this

ynotdraw:
> Code LGTM, but like Nicole, I couldn't find where to test this

Up in the PR description there's a "how to test" section. Should I do something different for next time?

danwenzel:
Maybe the link is wrong?  Or I'm missing something... :)

The link brings me to another github comment, which links me here: https://1b06c357.ember-toucan-core.pages.dev/

But that's just the main docs page.  Which component do I click into?

ynotdraw:
> Maybe the link is wrong?  Or I'm missing something... :)
> 
> 
> 
> The link brings me to another github comment, which links me here: https://1b06c357.ember-toucan-core.pages.dev/
> 
> 
> 
> But that's just the main docs page.  Which component do I click into?

Ah. Yeah correct. I guess maybe I should've outlined each component I touched and to review, although I guess you can also see them in the changed files.

The spacing defined in the PR description is what we want though. This'll all maybe be a bit easier when we integrate with Percy as it's all manual at the moment

danwenzel:
Ah, yeah, I should've caught that.  

Ok, so the more I'm looking at this... Are we sure we want to have both focus and error outlines at the same time?  It looks a bit strange to me, plus, if an input is focused, I'd prefer it not be red.  I tend to not like there being excessive reminder that I've got something wrong... But that's just my opinion :)

But another thing, leaving room for these two outlines add quite a bit of space between the label and the input, where there isn't that outer outline present:

image

ynotdraw:
> Are we sure we want to have both focus and error outlines at the same time?
> But another thing, leaving room for these two outlines add quite a bit of space between the label and the input, where there isn't that outer outline present:

I can certainly pass along your feedback! We've been iterating on this for a while now over on https://github.com/CrowdStrike/ember-toucan-core/pull/65 with back and forth between design and providing preview URLs based on changes for them to test things out.  They also used these styles during our usability testing where we had really good results + feedback.

Things are still a bit early and could change as we get some folks using it, so nothing here is set in total stone! Yay for iterating!  We'll likely also remain in `beta` or `pre-1.0.0` until things settle and we land on a happy place (at least imo)

simonihmig (approved):

nicolechung (approved):

PR #81, opened 2023-03-17
feat: Add CheckboxGroupField component

### LINKED ISSUE #78: Add a CheckboxGroupField
Similar to the RadioGroupField component, we should also have a CheckboxGroupField to handle the selections.

ynotdraw:
## 🚀 Description
This PR introduces a `CheckboxGroupField`component. 

Closes #78 

---

## 🔬 How to Test

- View the link at https://eb11f0be.ember-toucan-core.pages.dev/docs/components/checkbox-group-field
- Verify you can click multiple checkboxes

---

Screenshot 2023-03-17 at 12 20 59 PM

https://user-images.githubusercontent.com/8069555/225961503-cc83556f-958d-4324-9645-bdc9c7d33ef1.mov

### DIFF
--- a/docs/components/checkbox-group-field/demo/base-demo.md
+++ b/docs/components/checkbox-group-field/demo/base-demo.md
@@ -1,0 +1,38 @@
+```hbs template
+<Form::CheckboxGroupField
+  @label='Label'
+  @hint='Extra information about the field'
+  @name='options'
+  @value={{this.groupValue}}
+  @onChange={{this.updateValue}}
+  as |group|
+>
+  <group.CheckboxField @label='Option 1' @option='option-1' />
+  <group.CheckboxField
+    @label='Option 2'
+    @option='option-2'
+    @hint='Extra information about the radio'
+  />
+  <group.CheckboxField @label='Option 3' @option='option-3' />
+  <group.CheckboxField
+    @label='Option 4'
+    @option='option-4'
+    @hint='Extra information about the radio'
+  />
+</Form::CheckboxGroupField>
+```
+
+```js component
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class extends Component {
+  @tracked groupValue = [];
+
+  @action
+  updateValue(value) {
+    this.groupValue = value;
+  }
+}
+```
--- a/ember-toucan-core/src/components/form/checkbox-field.hbs
+++ b/ember-toucan-core/src/components/form/checkbox-field.hbs
@@ -13,9 +13,11 @@
           aria-invalid={{if @error "true"}}
           aria-describedby="{{if @error field.errorId}}"
           id={{field.id}}
+          name={{@name}}
+          value={{@option}}
           @isDisabled={{@isDisabled}}
           @isIndeterminate={{@isIndeterminate}}
-          @value={{@value}}
+          @value={{this.isChecked}}
           @onChange={{@onChange}}
           ...attributes
         />
--- a/ember-toucan-core/src/components/form/checkbox-field.ts
+++ b/ember-toucan-core/src/components/form/checkbox-field.ts
@@ -78,6 +78,12 @@ export default class ToucanFormCheckboxFieldComponent extends Component<ToucanFo
     args: ToucanFormCheckboxFieldComponentSignature['Args']
   ) {
     assert('A "@label" argument is required', args.label);
+
+    assert(
+      'Both "@option" and "@value" arguments were supplied. "@option" is reserved for being used in a CheckboxGroupField to specify the value attribute, while "@value" sets the checked state of the checkbox. Please use either "@option" or "@value", but not both.',
+      !(args.value && args.option)
+    );
+
     super(owner, args);
   }
 
--- a/ember-toucan-core/src/components/form/checkbox-field.ts
+++ b/ember-toucan-core/src/components/form/checkbox-field.ts
@@ -92,7 +92,7 @@ export default class ToucanFormCheckboxFieldComponent extends Component<ToucanFo
       return this.args.value;
     }
 
-    if (!this.args.selectedValues || this.args.selectedValues?.length === 0) {
+    if (!this.args.selectedValues || this.args.selectedValues.length === 0) {
       return false;
     }
 
--- a/ember-toucan-core/src/components/form/checkbox-field.ts
+++ b/ember-toucan-core/src/components/form/checkbox-field.ts
@@ -62,6 +62,8 @@ export interface ToucanFormCheckboxFieldComponentSignature {
      * This component argument is used to determine if the underlying checkbox is checked.
      * When the `selectedValues` array contains the `option`, the checkbox will be checked.
      * This should only be used for checkbox groups via CheckboxGroupField.
+     *
+     * @internal
      */
     selectedValues?: Array<string>;
 
--- a/ember-toucan-core/src/components/form/checkbox-field.ts
+++ b/ember-toucan-core/src/components/form/checkbox-field.ts
@@ -34,14 +34,36 @@ export interface ToucanFormCheckboxFieldComponentSignature {
     label: string;
 
     /**
+     * Sets the name attribute of the checkbox. A string specifying a name for the input control. This name is submitted along with the control's value when the form data is submitted.
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name
+     */
+    name?: string;
+
+    /**
      * The function called when the element is clicked.
      */
     onChange?: ToucanFormCheckboxControlComponentSignature['Args']['onChange'];
 
     /**
+     * The option argument gets mapped to the underlying value attribute of the checkbox. This should only be used
+     * for checkbox groups via CheckboxGroupField.
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes
+     */
+    option?: string;
+
+    /**
      * A test selector for targeting the root element of the field. In this case, the wrapping div element.
      */
     rootTestSelector?: string;
+
+    /**
+     * This component argument is used to determine if the underlying checkbox is checked.
+     * When the `selectedValues` array contains the `option`, the checkbox will be checked.
+     * This should only be used for checkbox groups via CheckboxGroupField.
+     */
+    selectedValues?: Array<string>;
 
     /**
      * Sets the checked state of the checkbox.
@@ -57,5 +79,17 @@ export default class ToucanFormCheckboxFieldComponent extends Component<ToucanFo
   ) {
     assert('A "@label" argument is required', args.label);
     super(owner, args);
+  }
+
+  get isChecked() {
+    if (!this.args?.option) {
+      return this.args.value;
+    }
+
+    if (!this.args.selectedValues || this.args.selectedValues?.length === 0) {
+      return false;
+    }
+
+    return this.args.selectedValues?.includes(this.args.option);
   }
 }
--- a/ember-toucan-core/src/components/form/checkbox-group-field.ts
+++ b/ember-toucan-core/src/components/form/checkbox-group-field.ts
@@ -1,0 +1,82 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+import CheckboxFieldComponent from './checkbox-field';
+
+import type { WithBoundArgs } from '@glint/template';
+
+export interface ToucanFormCheckboxGroupFieldComponentSignature {
+  Element: HTMLFieldSetElement;
+  Args: {
+    /**
+     * Provide a string to this argument to render an error message and apply error styling to the Field.
+     */
+    error?: string;
+
+    /**
+     * Provide a string to this argument to render a hint message to help describe the control.
+     */
+    hint?: string;
+
+    /**
+     * Sets the disabled attribute on the fieldset.
+     */
+    isDisabled?: boolean;
+
+    /**
+     * Provide a string to this argument to render inside of the label tag.
+     */
+    label: string;
+
+    /**
+     * Sets the name attribute of the checkboxes. A string specifying a name for the input control. This name is submitted along with the control's value when the form data is submitted.
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name
+     */
+    name: string;
+
+    /**
+     * The function called when a checkbox element is clicked.
+     */
+    onChange?: (selectedValues: Array<string>, e: Event | InputEvent) => void;
+
+    /**
+     * A test selector for targeting the root element of the field. In this case, the wrapping div element.
+     */
+    rootTestSelector?: string;
+
+    /**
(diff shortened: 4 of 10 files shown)


ember-toucan-core/src/components/form/checkbox-field.hbs
```diff
@@ -13,9 +13,11 @@
           aria-invalid={{if @error "true"}}
           aria-describedby="{{if @error field.errorId}}"
           id={{field.id}}
+          name={{@name}}
+          value={{@option}}
```
ynotdraw, line 17: More info for this at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes, but for a group of checkboxes, to differentiate one from the other, the `value` attribute is used directly.

In our case, `@value` is for the checked-state of the checkbox.  `@option` is used to define "this is a checkbox option" and on-click we can know which checkbox was checked or unchecked.  This is a little confusing as we now have a `value` attribute and a `@value` component argument, so I tried to add documentation below to assist.

ember-toucan-core/src/components/form/checkbox-field.ts
```diff
@@ -58,4 +80,16 @@ export default class ToucanFormCheckboxFieldComponent extends Component<ToucanFo
     assert('A "@label" argument is required', args.label);
     super(owner, args);
   }
+
+  get isChecked() {
+    if (!this.args?.option) {
+      return this.args.value;
+    }
```
ynotdraw, line 87: This essentially controls the case where there's a single checkbox in use.  Single checkboxes do not use `@option`, so we only need to respect `@value`.

ember-toucan-core/src/components/form/checkbox-field.ts
```diff
@@ -58,4 +80,16 @@ export default class ToucanFormCheckboxFieldComponent extends Component<ToucanFo
...
+
+    if (!this.args.selectedValues || this.args.selectedValues?.length === 0) {
+      return false;
+    }
+
+    return this.args.selectedValues?.includes(this.args.option);
```
ynotdraw, line 93: This is where things get a little trickier in a checkbox group.  Essentially we need to check if the `@option` is in the `@selectedValues` array to determine if this particular checkbox is checked or not.

ember-toucan-core/src/components/form/checkbox-group-field.ts
```diff
@@ -0,0 +1,82 @@
...
+  CheckboxFieldComponent = CheckboxFieldComponent;
+
+  @action
+  handleInput(_: boolean, e: Event | InputEvent): void {
+    let value = this.args.value ? [...this.args.value] : [];
+
+    let target = e.target as HTMLInputElement;
+
+    if (value?.includes(target.value)) {
+      value = value.filter((val) => val !== target.value);
```
ynotdraw, line 75: Some code golf could happen to make this a bit better, but:

1. `value` should be a pretty small array (we don't want people making 10 checkbox items, for example, instead they should probably use multi-select or something)
2. This is a one-liner

docs/components/checkbox-group-field/demo/base-demo.md
```diff
@@ -0,0 +1,38 @@
...
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class extends Component {
+  @tracked groupValue = [];
+
+  @action
+  updateValue(value) {
+    this.groupValue = value;
```
ynotdraw, line 35: Docs on this below, but `value` here will be an array of strings that'll align with the `@option` clicked.  Consider the following:

- A user clicks "Option 1"
- `onChange` is called, and the `value` here will be `['option-1']` (note: we use the `@option` provided to tell the user which one was clicked!
- User clicks "Option 3"
- `onChange` is now called, and the `value` here will be `['option-1', 'option-3']`
- User clicks "Option 1" again, unchecking the checkbox
- `onChange` will be `['option-3']`

ember-toucan-core/src/components/form/checkbox-group-field.ts
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+      value.push(target.value);
+    }
+
+    this.args.onChange?.(value, e);
```
ynotdraw, line 80: Simon and I talked about maybe making the first argument to `onChange` a `Map`/POJO that could look like:

```
{ 'option-1': true, 'option-2': false, 'option-3: false }
```

After looking at existing uses, it looks like we expect this to be an array of the selected items (as done in this PR!).  Due to that, we punted on this; however, as we start integrating it more we may switch to something like this if it makes more sense.

test-app/tests/integration/components/checkbox-group-field-test.gts
```diff
@@ -0,0 +1,234 @@
...
+  test('it sets the `value` attribute of the yielded CheckboxFields based on the provided `@option` argument', async function (assert) {
+    await render(<template>
+      <CheckboxGroupField
+        @label="Label"
+        @name="group"
+        data-group-field
+        as |group|
+      >
+        <group.CheckboxField
+          @label="option-1"
```
nicolechung, line 59: Nit, suggestion: can we make the `@label` versus `@option` different for readability? That way I can tell `option-1` is coming from `@option`.

test-app/tests/integration/components/checkbox-group-field-test.gts
```diff
@@ -0,0 +1,234 @@
...
+        />
+        <group.CheckboxField
+          @label="option-2"
+          @option="option-2"
+          data-checkbox-2
+        />
+      </CheckboxGroupField>
+    </template>);
+
+    assert.dom('[data-checkbox-1]').hasAttribute('name', 'group');
```
nicolechung, line 96: Suggestion: similar thing here...can we change the `@name` to something other than `group` because it can be confused with `|group|` when scanning the code.

test-app/tests/integration/components/checkbox-group-field-test.gts
```diff
@@ -0,0 +1,234 @@
...
+        <group.CheckboxField
+          @label="option-2"
+          @option="option-2"
+          data-checkbox-2
+        />
+      </CheckboxGroupField>
+    </template>);
+
+    assert.dom('[data-checkbox-1]').isNotChecked();
+    assert.dom('[data-checkbox-2]').isChecked();
```
nicolechung, line 140: Nice!

ember-toucan-core/src/components/form/checkbox-group-field.ts
nicolechung: Would we need to modify `ember-headless-form` to yield the value/values?

https://github.com/CrowdStrike/ember-headless-form/blob/e6398198dfa1a1d9d51cf439c7eb67fc85d1555d/docs/usage/controls.md#radio-group

It looks like it yields the rawErrors now but not the rawValue if I am reading this correctly: https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/components/headless-form.hbs#L31

ember-toucan-core/src/components/form/checkbox-group-field.ts
nicolechung: Seems pretty readable to me with `includes` and `filter`

ember-toucan-core/src/components/form/checkbox-field.hbs
nicolechung: With this part: 
```hbs
value={{@option}}
```

1. Aren't we coupling the checkbox-field to `checkbox-group-field`. Do we want to do this? What if people want to have a single checkbox, and not a group.
2. `@option` seems optional, so perhaps this should be wrapped in an `if` statement if there is no option provided.

ember-toucan-core/src/components/form/checkbox-field.hbs
```diff
@@ -13,9 +13,11 @@
...
           @isDisabled={{@isDisabled}}
           @isIndeterminate={{@isIndeterminate}}
-          @value={{@value}}
+          @value={{this.isChecked}}
```
nicolechung, line 20: So:
1. `@value` determines if the checkbox is checked? https://github.com/CrowdStrike/ember-toucan-core/blob/main/ember-toucan-core/src/components/form/controls/checkbox.ts#L36
2. `value` (no `@`) determines the "option" in a group of options?

ember-toucan-core/src/components/form/checkbox-field.ts
nicolechung: Ahhh....okay I'm sort of getting this now. 

Suggestion: I think the checkbox-field docs could be updated to reflect this.

nicolechung: Is there a way to prevent the user from passing `@option` if the parent is not a checkbox field? 

(I don't actually know if there is a way to do this)

nicolechung (approved):

ember-toucan-core/src/components/form/checkbox-field.hbs
nicolechung: NVM I re-read what you wrote and wrote below...

ember-toucan-core/src/components/form/checkbox-field.ts
ynotdraw: So I was playing with discriminate unions in TS as I've used those before, but I couldn't quite get what I wanted with how the contextual components work. I think I may be able to use some asserts though to help prompt people to use the correct arguments. Thanks for the suggestion! I'll get something up tomorrow AM

test-app/tests/integration/components/checkbox-group-field-test.gts
ynotdraw: Went ahead and updated with https://github.com/CrowdStrike/ember-toucan-core/pull/81/commits/abb6338f6bfbc957e4335850e0d2bca591caab96 - thanks!

test-app/tests/integration/components/checkbox-group-field-test.gts
ynotdraw: Went ahead and updated with https://github.com/CrowdStrike/ember-toucan-core/pull/81/commits/abb6338f6bfbc957e4335850e0d2bca591caab96 - thanks!

ember-toucan-core/src/components/form/checkbox-field.hbs
ynotdraw: Yup yup, more info over on https://github.com/CrowdStrike/ember-toucan-core/pull/81#discussion_r1140586395!

ember-toucan-core/src/components/form/checkbox-field.ts
ynotdraw: Okay so still no luck with TS, but I did add an assertion that essentially ensures folks either provide `@option` or `@value`, but not **both**.

https://github.com/CrowdStrike/ember-toucan-core/pull/81/commits/78e90b8a56783dfbbf6d45256c95cec9bc2b4407

Let me know what you think!

nicolechung: This looks great!

ember-toucan-core/src/components/form/checkbox-group-field.ts
```diff
@@ -0,0 +1,77 @@
...
+     * The currently selected checkbox elements. The elements with the matching `@option` / underlying value attribute will be checked.
+     */
+    value?: Array<string>;
+  };
+  Blocks: {
+    default: [
+      {
+        CheckboxField: WithBoundArgs<
+          typeof CheckboxFieldComponent,
+          'isDisabled' | 'name' | 'value'
```
simonihmig, line 53: Is this correct? According to the template, it should be [these four arguments](https://github.com/CrowdStrike/ember-toucan-core/pull/81/files#diff-f513e3b13a120c52fd2a93f32c4f56af79da3ded446cbb4ebae0a1c2b9a0717aR13-R16)?

Wonder why Glint seems happy though!? 🤔

simonihmig (approved):
Looks good, just some minor remarks! :+1:

docs/components/checkbox-group-field/demo/base-demo.md
```diff
@@ -0,0 +1,38 @@
...
+  @hint='Extra information about the field'
+  @name='options'
+  @value={{this.groupValue}}
+  @onChange={{this.updateValue}}
+  as |group|
+>
+  <group.CheckboxField @label='Option 1' @option='option-1' />
+  <group.CheckboxField
+    @label='Option 2'
+    @option='option-2'
```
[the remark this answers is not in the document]
ynotdraw: So the current `CheckboxField` component uses `@value` to denote if it is checked or not:

- https://github.com/CrowdStrike/ember-toucan-core/blob/main/ember-toucan-core/src/components/form/checkbox-field.hbs#L18
- https://github.com/CrowdStrike/ember-toucan-core/blob/main/ember-toucan-core/src/components/form/controls/checkbox.hbs#L6
- https://github.com/CrowdStrike/ember-toucan-core/blob/main/ember-toucan-core/src/components/form/controls/checkbox.ts#L36

So to differentiate between a checkbox group option and a checkbox "checked state", I opted for `@option` here.  We also haves some prior art for this (aka existing code) that follows the same component API pattern.

Brings up a good question though.  Maybe it'd be better to make the underlying checkbox control use a `@isChecked` instead of `@value`?  Then we can update CheckboxGroupField here to `@value` for each individual checkbox?  That may be a cleaner API.  What do you think?


```hbs
<Form::CheckboxGroupField
  @label='Label'
  @hint='Extra information about the field'
  @name='options'
  @value={{this.groupValue}}
  @onChange={{this.updateValue}}
  as |group|
>
  <group.CheckboxField @label='Option 1' @value='option-1' />
  <group.CheckboxField
    @label='Option 2'
    @value='option-2'
    @hint='Extra information about the radio'
  />
  <group.CheckboxField @label='Option 3' @option='option-3' />
  <group.CheckboxField
    @label='Option 4'
    @value='option-4'
    @hint='Extra information about the radio'
  />
</Form::CheckboxGroupField>
```


And then the checkbox-field component:

```hbs
<Form::CheckboxField
  @label='Label'
  {{! currently this is "@value" instead }}
  @isChecked={{this.value}}
  @onChange={{this.updateValue}}
/>
```

```js
import { tracked } from '@glimmer/tracking';

export default class extends Component {
  @tracked value = false;

  @action
  updateValue(value) {
    this.value = value;
  }
}
```

---

If above makes sense and we want to do it, I may do that as a separate follow up PR that's focused on the component argument renaming and what not so that this PR doesn't continue to grow in scope if that's okay with you.

ember-toucan-core/src/components/form/checkbox-group-field.ts
ynotdraw: Oh my word, excellent eyes!  I forgot to update these when iterating.  That's super interesting.  I do wonder why Glint didn't complain.  Think this is Glint issue worthy?

ynotdraw: Sounds excellent to me! I'm going to go ahead and merge this as is if that's okay with you, but we should chat about your findings on this and maybe https://github.com/CrowdStrike/ember-toucan-core/pull/81#discussion_r1143491402 as follow ups later this week if you have some time! 🙇

Issue #91, opened 2023-03-22
FileInputField, handle multiple in the control file input instead


nicolechung:
@ynotdraw:

I wonder if we can remove multiple here and instead rely on https://github.com/CrowdStrike/ember-toucan-core/pull/64/files#diff-262ffa41ec148d582dd8e475dc10b89c5a53f370ddc56ed991326fa6d90e6bffR33? And maybe accept should get moved to the control as well, so that by default the control file input also defaults to *?

PR #94, opened 2023-03-22
fix: FileInputField, remove multiple helper

nicolechung:
The multiple helper in FileInputField already exists in controls/file-input.

Deleting this and passing along the @multiple arg works and is simpler.

Fixes: https://github.com/CrowdStrike/ember-toucan-core/issues/91

- [x] move `accept` over as well, see comment: https://github.com/CrowdStrike/ember-toucan-core/pull/64/files#r1144078795
- [x] add @multiple to the docs

### DIFF
--- a/ember-toucan-core/src/components/form/file-input-field.ts
+++ b/ember-toucan-core/src/components/form/file-input-field.ts
@@ -50,10 +50,6 @@ export default class ToucanFormFileInputFieldComponent extends Component<ToucanF
     return this.args.accept ?? '*';
   }
 
-  get multiple() {
-    return this.args.multiple ?? true;
-  }
-
   get hasError() {
     return Boolean(this.args?.error);
   }
--- a/ember-toucan-core/src/components/form/file-input-field.ts
+++ b/ember-toucan-core/src/components/form/file-input-field.ts
@@ -46,10 +46,6 @@ export default class ToucanFormFileInputFieldComponent extends Component<ToucanF
     super(owner, args);
   }
 
-  get accept() {
-    return this.args.accept ?? '*';
-  }
-
   get hasError() {
     return Boolean(this.args?.error);
   }
--- a/docs/components/file-input-field/index.md
+++ b/docs/components/file-input-field/index.md
@@ -2,6 +2,11 @@
 
 Provides an underlying `<input type="file">` element building on top of the Field component.
 
+## Accepts
+
+Optional.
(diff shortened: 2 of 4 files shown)


ynotdraw (approved):

ember-toucan-core/src/components/form/file-input-field.ts
```diff
@@ -50,10 +50,6 @@ export default class ToucanFormFileInputFieldComponent extends Component<ToucanF
     return this.args.accept ?? '*';
   }
 
-  get multiple() {
-    return this.args.multiple ?? true;
```
danwenzel, line 54: Just making sure: With this change, if they don't pass `@multiple`, it'll now end up `false`, whereas before, it would have defaulted to `true`.  That's what we want?

danwenzel: That logic looks different, though...

`this.args.multiple ?? true;` means that it'll be `true` if they don't pass `@multiple`
`this.args.multiple ? true : undefined;` means that it'll be `undefined` if they don't pass `@multiple`

nicolechung: Oh yeah this is coming back to me.

1. We want either `multiple` or  NO multiple attribute whatsoever, so we do want `undefined` NOT `multiple=false`
2. `Multiple=false` still weirdly allows users to select multiple files to upload which is NOT what we want.

Can verify that over here: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_multiple
1. try `multiple=false` 
2. try removing `multiple` altogether
3. Notice #1 and #2 do not give the same results. 

We validate that the `multiple` attribute is not there when `@multiple={{false}}` is passed in this test: https://github.com/CrowdStrike/ember-toucan-core/blob/main/test-app/tests/integration/components/file-input-field-test.gts#L331

So if all tests are passing in this PR, then we are getting the intended result (no `multiple` arg)!

danwenzel: cool, that all makes sense.  So now, if the `@multiple` arg is not provided at all, the `multiple` HTML attribute will be omitted altogether, so the input will _not_ allow multiple selections.

Can we add `@multiple` to the documentation?

danwenzel (approved):

ember-toucan-core/src/components/form/file-input-field.ts
nicolechung: Great idea! Let me add this...to the right doc

PR #96, opened 2023-03-23
fix: reinstall minus lockfile and pnpm dedupe

nicolechung:
With regards to: https://github.com/embroider-build/embroider/issues/1378


nicolechung:
Closed as we are moving back to a previous version of pnpm to fix this.

PR #100, opened 2023-03-23
style: FileInputField, list, add styles

nicolechung:
## Description 🚀 

- [x] style the list item more like the `input` in  the text input field
- [x] **test update**: for `input` without a `multiple attribute` (i.e. single file uploads) we want to be able to change the `triggerText` to indicate that this is a `replace` not just adding more files. I updated the test to reflect this.

Note: this does not change the existing markup of the file input field, that will be done in followup work.

## Before
Screenshot 2023-03-23 at 4 19 23 PM

## After
Screenshot 2023-03-27 at 1 06 22 PM

### DIFF
--- a/ember-toucan-core/src/components/form/file-input/delete-button.hbs
+++ b/ember-toucan-core/src/components/form/file-input/delete-button.hbs
@@ -3,7 +3,7 @@
   type="button"
   aria-label={{@deleteLabel}}
   {{! TODO: remove reset-styles https://github.com/CrowdStrike/ember-oss-docs/issues/17 }}
-  class="reset-styles flex-none bg-transparent"
+  class="reset-styles mr-2 flex-none bg-transparent"
   {{on "click" (fn @onDelete @file)}}
   ...attributes
 >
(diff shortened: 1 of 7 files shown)


ember-toucan-core/src/components/form/file-input/delete-button.hbs
```diff
@@ -3,7 +3,7 @@
   type="button"
   aria-label={{@deleteLabel}}
   {{! TODO: remove reset-styles https://github.com/CrowdStrike/ember-oss-docs/issues/17 }}
-  class="reset-styles flex-none bg-transparent"
+  class="reset-styles mr-2 flex-none bg-transparent"
```
ynotdraw, line 6: non-blocking question: Should this class be moved onto the consumer at  `ember-toucan-core/src/components/form/file-input/list-item.hbs`:

```hbs
<Form::FileInput::DeleteButton class="mr-2"
```

so that there isn't built-in margin to the button but rather in the "wrapping" component?

ynotdraw (approved):

simonihmig (approved):

PR #108, opened 2023-03-27
File input: redo input to button style, move triggerText to the left

nicolechung:
## 🚀 Description
The design for the FileInputField has changed. The input field should now look more like our secondary button variant, not like a regular text input field.

**Note**: while you can style [file input buttons with tailwind](https://tailwindcss.com/docs/hover-focus-and-other-states#file-input-buttons), the actual file input button is a shadow-dom element, and (as far as I know) there is not way to insert an SVG (which we use for the "File upload" icon).

To get around this, I styled a `div` inside of the label LIKE a button. We still want to have a separate label and hint, and so the `label` tag still wraps around everything else in order to mimic the button functionality / hit zone we are after. 

- [x] changed `flex` to `inline-flex` to reduce the width of the clickable area to just the "button" width
- [x] hid the actual `<input>` with `sr-only`
- [x] changed the `div` that surrounds the input to be styled like our secondary button
- [x] swapped the position of the triggerText ("Select Files") and the icon
- [x] added margin-left to the icon to space it away from the text
- [x] changed `Select Files` to `Browse Files` to match the design

---

## 🔬 How to Test

1. tests should pass

---

## Before
Screenshot 2023-03-27 at 12 05 39 PM

## After
Screenshot 2023-03-27 at 12 05 14 PM


test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -300,7 +297,9 @@ module('Integration | Component | FileInputField', function (hooks) {
 
     // Verify files are there
     assert.dom('[data-files]').exists();
-    await triggerEvent('button', 'click');
+    await triggerEvent('[data-delete-file]', 'click');
+
+
```
ynotdraw, line 302: Super nit: I think VSCode prettier may complain about this being two returns rather than 1, but may be okay?

ynotdraw (approved):
:shipit:

ember-toucan-core/src/components/form/file-input-field.hbs
```diff
@@ -24,7 +24,24 @@
           ...attributes
         />
       </field.Control>
-
+      <Button @variant="secondary" @onClick={{(fn this.handleChange field)}}>
```
simonihmig, line 27: Shouldn't this be also inside of `<field.Control>`?

Also, and this is really weird, looking at the  button's [template](https://github.com/CrowdStrike/ember-toucan-core/blob/main/ember-toucan-core/src/components/button.hbs), this should be a literal `<button>`. But why am I seeing this as a div in Chrome!? 🤯

image

simonihmig: Oh, wait, taking the URL from the preview link (https://678302d9.ember-toucan-core.pages.dev/docs/components/file-input-field), the button is actually a button. 

But it is not on the URL that was in the PR description! 🤔

simonihmig (approved):
Approving already, but please look into...

> Shouldn't this be also inside of <field.Control>?

ember-toucan-core/src/components/form/file-input-field.hbs
nicolechung: I was having issues with Github actions and publishing the docs yesterday (first Github Actions had problems, then I had a typo in one of my base demos) so maybe that's why....

test-app/tests/integration/components/file-input-field-test.gts
nicolechung: I removed it manually...tried opening up VS Code to format this, but then got red lines everywhere...

ember-toucan-core/src/components/form/file-input-field.hbs
nicolechung: Moved the button into control here: https://github.com/CrowdStrike/ember-toucan-core/pull/108/commits/3bb6d0349bfcfd280691b30045eaefb327736483

PR #110, opened 2023-03-27
Add labels and hint blocks to fields

nicolechung:
## 🚀 Description

Added `label` and `hint` named blocks to field components.

- [x] added: `label` and `hint` named blocks to field components
- [x] added: `label` and `hint` to fieldset, radio group field, and checkbox group field components
- [x] ⛔️ removed: `aria-invalid` from the checkbox field group, due to this lint rule: https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-unsupported-role-attributes.md
  - [x] removed: the related checkbox group field test for `aria-invalid` 
 - [x] 👉 update: test titles to `Component | Fields | { Component Name }`
 - [x] added: use `@label` or `:label` but not both at the same time
 - [x] added: use `@hint` or `:hint` but not both at the same time
 - [x] added: if `@label` and `:label` are missing, error because a label is required (either as a block or arg)
 - [x] removed: blocks from all the base-demo files
 - [x] added: tests for each field component
   - [x] it can render a `:hint` and `:label` block
   - [x] it throws an assertion error if no `@label` is provided
   - [x] it throws an assertion error if both `@label` and `:label` are provided
  - [x] docs: added @arg versus `:named-block` implementation for `Label` and `Hint`
  - [ ] delete `Fieldset` and related tests   

## 🔬 How to Test

1. Check out all the Field components and make sure `:hint` and `:label` have been added: https://dc754599.ember-toucan-core.pages.dev/docs/components/file-input-field

---

Screenshot 2023-03-31 at 2 40 45 PM

### DIFF
--- a/.changeset/famous-foxes-joke.md
+++ b/.changeset/famous-foxes-joke.md
@@ -1,0 +1,6 @@
+---
+'@crowdstrike/ember-toucan-core': minor
+---
+
+- Added label and hint named blocks to all the Field components.
+- Removed Fieldset.


docs/components/checkbox-field/demo/base-demo.md
```diff
@@ -3,7 +3,10 @@
   @label='Label'
   @isChecked={{this.isChecked}}
   @onChange={{this.handleChange}}
-/>
+>
+  <:label>extra label info</:label>
```
nicolechung, line 7: Was not feeling super imaginative here...if anyone has any suggestions of what to put in the blocks I will do a sweep and update the content in a single commit!

ynotdraw:
Sorry for busting your PR again, but now I'm done moving things/renaming things!

nicolechung:
> Sorry for busting your PR again, but now I'm done moving things/renaming things!

Nice to see things getting reorganized tho!

ember-toucan-core/src/components/form/fields/checkbox-group.hbs
```diff
@@ -1,20 +1,41 @@
...
+  <fieldset
+    aria-describedby="{{if @error field.errorId}} {{if @hint field.hintId}}"
+    disabled={{@isDisabled}}
+    {{! attribute aria-invalid is not supported by the element fieldset with the implicit role of group }}
+    ...attributes
+  >
+    <legend
+      class="type-md-tight text-body-and-labels block"
+      data-label
+    >{{@label}} {{yield to="label"}}</legend>
```
simonihmig, line 11: General question: do we really want to render both the argument and the block, in case both are provided? This is not wrong per se, but maybe confusing? We _could_ instead also give one a higher priority, and render only that:

Arg has priority:
```hbs
{{#if @label}}
  {{@label}}
{{else}}
  {{yield to="label"}}
{{/if}}
```

or block has priority:
```hbs
{{#if (has-block "label")}}
  {{yield to="label"}}
{{else}}
  {{@label}}
{{/if}}
```

What do you think?

ember-toucan-core/src/components/form/fields/checkbox-group.hbs
```diff
@@ -1,20 +1,41 @@
...
-        onChange=this.handleInput
-        isDisabled=@isDisabled
-        selectedValues=@value
-      )
-    )
-  }}
-</Form::Fieldset>
\ No newline at end of file
+<Form::Field as |field|>
+  <fieldset
```
simonihmig, line 2: For my understanding: what was the reason to not use `<Form::Fieldset>` anymore?

ember-toucan-core/src/components/form/fields/checkbox-group.hbs
nicolechung: Good question, I think rendering one or the other makes sense.

@ynotdraw  thoughts?

ynotdraw: I agree on choosing one or the other but not both! I'm leaning towards block having priority for no real reason other but have 0 strong opinions.

docs/components/checkbox-field/demo/base-demo.md
ynotdraw: You could maybe do something like what we are expecting to do - add an icon here (with no tooltip for now):

```html
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" stroke="currentColor" viewBox="0 0 24 24"><path d="M12 3a9 9 0 11-6.364 2.636A8.972 8.972 0 0112 3zm0 4.7v5.2m0 3.39v.01" fill="none" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></path></svg>
```

docs/components/checkbox-field/demo/base-demo.md
```diff
@@ -3,7 +3,10 @@
...
+  <:hint>extra hint info</:hint>
```
ynotdraw, line 8: For this you could maybe add a `<a href="https://www.crowdstrike.com">Learn more</a>` with some additional styling to make it look like a Toucan link

docs/components/checkbox-group-field/demo/base-demo.md
```diff
@@ -5,20 +5,23 @@
...
+    />
+    <group.CheckboxField @label='Option 3' @value='option-3' />
+    <group.CheckboxField
+      @label='Option 4'
+      @hint='Extra information about the radio'
+      @value='option-4'
+    />
+  </:default>
+  <:label>extra label info</:label>
+  <:hint>extra hint info</:hint>
```
ynotdraw, line 24: I'm wondering:

Should the named-block options go under a new section + under the UI states sections instead?  I feel like they could use some extra documentation and that'd be a good way to show usage + what they look like visually.

packages/ember-toucan-core/src/components/form/fields/checkbox-group.hbs
```diff
@@ -1,20 +1,41 @@
...
-      )
-    )
-  }}
-</Form::Fieldset>
\ No newline at end of file
+<Form::Field as |field|>
+  <fieldset
+    aria-describedby="{{if @error field.errorId}} {{if @hint field.hintId}}"
+    disabled={{@isDisabled}}
+    {{! attribute aria-invalid is not supported by the element fieldset with the implicit role of group }}
```
ynotdraw, line 5: Ah - so no more `aria-invalid={{if @error "true"}}` due to this?

packages/ember-toucan-core/src/components/form/fields/checkbox.ts
```diff
@@ -73,6 +73,10 @@ export interface ToucanFormCheckboxFieldComponentSignature {
      */
     value?: string;
   };
+  Blocks: {
+    label: unknown;
+    hint: unknown;
```
ynotdraw, line 78: `unknown` here - should these be `[]`?

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -100,6 +100,31 @@ module('Integration | Component | FileInputField', function (hooks) {
...
+      >
+      <:label>Extra label content</:label>
+      <:hint>Extra hint content</:hint>
+      </FileInputField>
+    </template>);
+
+    // For the file input field component, the only aria-describedby
+    // value should be the errorId.  This is due to the way the component
+    // is structured, where the label+hint are rendered inside of the
+    // wrapping <label> element
```
ynotdraw, line 121: Looks like maybe copy-paste comment that's not needed here?

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -110,6 +110,26 @@ module('Integration | Component | InputField', function (hooks) {
...
+      <:label>Extra label content</:label>
+      <:hint>Extra hint content</:hint>
+      </InputField>
+    </template>);
+
+    const label = '[data-label]';
+    const hint = '[data-hint]';
+
+    assert.dom(label).hasText('Label Extra label content'); 
+    assert.dom(hint).hasText('Hint text visible here Extra hint content');
```
ynotdraw, line 130: nit: Instead of doing a `find` can these be the following to match the other tests above?

```suggestion
    assert.dom('[data-hint]').hasText('Hint text Extra hint content');
    assert.dom('[data-label]').hasText('Label Extra label content');
```

test-app/tests/integration/components/textarea-field-test.gts
```diff
@@ -55,6 +55,21 @@ module('Integration | Component | TextareaField', function (hooks) {
...
+        <:label>Extra label content</:label>
+        <:hint>Extra hint content</:hint>
+      </TextareaField>
+    </template>);
+
+    let hint = find('[data-hint]');
+    let label = find('[data-label]');
+
+    assert.dom(hint).hasText('Hint text Extra hint content');
+    assert.dom(label).hasText('Label Extra label content');
```
ynotdraw, line 70: nit: Instead of doing a `find` can these be the following to match the other tests above?

```suggestion
    assert.dom('[data-hint]').hasText('Hint text Extra hint content');
    assert.dom('[data-label]').hasText('Label Extra label content');
```

packages/ember-toucan-core/src/components/form/fields/checkbox-group.hbs
nicolechung: Yeah I guess the linter (ember-template-lint?) doesn't catch this when we were adding it as `...attributes` on a component, but now that it can tell it's a group it doesn't want you to add `aria-invalid`?

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -110,6 +110,26 @@ module('Integration | Component | InputField', function (hooks) {
...
+      <:hint>Extra hint content</:hint>
+      </InputField>
+    </template>);
+
+    const label = '[data-label]';
+    const hint = '[data-hint]';
+
+    assert.dom(label).hasText('Label Extra label content'); 
+    assert.dom(hint).hasText('Hint text visible here Extra hint content'); 
+
```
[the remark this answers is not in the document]
nicolechung: Okay will give it a shot...whenever I open a gts file in my VS Code there are red lines everywhere so I just sort of assumed it wouldn't be formatting correctly.

docs/components/checkbox-field/demo/base-demo.md
nicolechung: oh wait...with these changes all these `:named-blocks` should be removed from these demos.

packages/ember-toucan-core/src/components/form/fields/checkbox.ts
nicolechung: THIS was missing tests so I missed this! Should be fixed in the next round because I added tests.

test-app/tests/integration/components/input-field-test.gts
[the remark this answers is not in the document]
nicolechung: Update - I think this did something! Will keep formatting `markdown` and `gts/gjs` files in VS Code from now on...

docs/components/checkbox-group-field/demo/base-demo.md
nicolechung: These `:label` and `:hint` blocks are removed from this demo now (they should not have been here).

I also added `:label` code snippets under each `## Label` section, same for `:hint`. I don't think they can go under UI states because folks can't see the code there...I guess they could see an icon for label or a link for hint? Seems more useful to have it under the `## Label` and `## Hint` sections tho (in terms of being able to copy those code snippets for consumers)

ember-toucan-core/src/components/form/fields/checkbox-group.hbs
nicolechung: We need to be able to put `{{yield to="label"}}` and `{{yield to="hint}}` inside of say, the block for `field.Label` and the `Fieldset` component does not yield these subcomponents.

packages/ember-toucan-core/src/components/form/fields/checkbox.hbs
```diff
@@ -25,14 +25,39 @@
...
-          >{{@label}}</span>
+          {{this.assert
+            (hash
+              blockExists=(has-block "label")
+              argName="label"
+              arg=@label
+              required=true
+            )
+          }}
+          {{#if (this.has (has-block "label") @label)}}
```
ynotdraw, line 36: I'm kind of wondering if we can combine these two helpers into one so that there's less logic here.  Essentially something like the following, without the need for `this.assert` plus the `{{#if (this.has` pieces:

```hbs
{{#if (this.assertBlockOrLabel (hash blockExists=(has-block "label") name="label" argument=@label) isRequired=true)}}
```

and the code for that being something like:

```js
// in the backing ts class

// I don't think this needs to go in `/helpers` as it's more of a shared function than a true Ember helper
import { assertBlockOrArgumentExists } from '../../../-private/assertBlockOrArgumentExists';
import type { AssertBlockOrArg }  from '../../../-private/assertBlockOrArgumentExists';

// ... Other stuff ...

assertBlockOrLabel = ({ blockExists, argument, name, isRequired }: AssertBlockOrArg) =>
    assertBlockOrArgumentExists({ blockExists, argument, name, isRequired });
```


```ts
import { assert } from '@ember/debug';

export type AssertBlockOrArg = {
  /**
   * The name of the argument that should be required.
   */
  argument?: string;
  /**
   * The name of the block or argument. Currently, the block and argument name must match.
   */
  name: string;
  /**
   * Does the provided block exist?
   */
  blockExists: boolean;
  /**
   * Should this assertion be required, or is it optional?
   */
  isRequired?: boolean;
};

export const assertBlockOrArgumentExists = ({
  argument,
  name,
  blockExists,
  isRequired
}: AssertBlockOrArg) => {
  if (isRequired) {
    assert(
      `Either a block "<:${name}>" or argument "@${name}" are required.`,
      Boolean(blockExists || argument)
    );
  }

  assert(
    `Either a block "<:${name}>" or argument "@${name}" may be provided, but not both`,
    !(blockExists && argument)
  );

 // By returning true here, our if block above should be satisfied
  return true;
};
```


I think by returning a boolean _after_ the assertions run, that should allow us to make this all 

```hbs
{{#if (this.assertBlockOrLabel (hash blockExists=(has-block "label") name="label" argument=@label) isRequired=true)}}
  <span
    class="type-md-tight block leading-4
      {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
    data-label
  >
    {{#if (has-block "label")}}
      {{yield to="label"}}
    {{else}}
      {{@label}}
    {{/if}}
  </span>
{{/if}}
```

packages/ember-toucan-core/src/components/form/fields/checkbox.hbs
```diff
@@ -25,14 +25,39 @@
...
+            {{this.assert
+              (hash blockExists=(has-block "hint") argName="hint" arg=@hint)
+            }}
+            <field.Hint data-hint>
+              {{#if (has-block "hint")}}
+                {{yield to="hint"}}
+              {{else}}
+                {{@hint}}
+              {{/if}}
+            </field.Hint>
```
ynotdraw, line 60: With above, I think this becomes a bit simpler as well:

```hbs
{{#if (this.assertBlockOrLabel (hash blockExists=(has-block "hint") name="hint" argument=@hint) isRequired=false)}}
  <field.Hint data-hint>
    {{#if (has-block "hint")}}
      {{yield to="hint"}}
    {{else}}
      {{@hint}}
    {{/if}}
  </field.Hint>
{{/if}}
```

packages/ember-toucan-core/src/-private/assert-block-or-argument-exists.ts
```diff
@@ -0,0 +1,39 @@
+import { assert } from '@ember/debug';
+
+export type AssertBlockOrArg = {
```
nicolechung, line 3: @ynotdraw  new singular function / not a helper.

docs/components/checkbox-group-field/index.md
```diff
@@ -8,11 +8,73 @@ Provides a checkbox group to be used within forms. It yields [CheckboxFields](./
 
 ## Label
 
-Provide a string to `@label` to render the text into the `<legend>` of the fieldset.
+Required.
+
+Use either `@label` or `:label`.
```
ynotdraw, line 13: nit: To maybe be "modern Ember" friendly or more friendly to folks new to Ember, I wonder if being a bit more verbose here would be helpful:

```suggestion
Use either the `@label` component argument or the `:label` named block.
```

docs/components/checkbox-group-field/index.md
```diff
@@ -8,11 +8,73 @@ Provides a checkbox group to be used within forms. It yields [CheckboxFields](./
...
+
+Provide a string to `@label` or components to `:label` to render the text into the `<legend>` of the fieldset.
```
ynotdraw, line 15: nit: the `:label` named block can really take _any_ content!

```suggestion
Provide a string to the `@label` component argument or content to the `:label` named block to render the into the `<legend>` of the fieldset.
```

docs/components/fieldset/demo/base-demo.md
```diff
@@ -3,10 +3,7 @@
   @label='Label'
   @hint='Extra information about the fieldset'
   @error='Error message'
->
-  <p class='text-body-and-labels text-xs m-0 italic'>~Fieldset components render
-    here!~</p>
-</Form::Fieldset>
```
ynotdraw, line 9: Any reason to axe this?

docs/components/fieldset/index.md
```diff
@@ -4,20 +4,56 @@ Fieldset is a component to aid in creating form components that require an under
 
 ## Label
 
-Provide a string to `@label` to render the text into the `<legend>` of the Fieldset. This is required.
+Required.
+
+Use either `@label` or `:label`.
+
+Provide a string to `@label` or components to `:label` to render the text into the `<legend>` of the Fieldset.
```
ynotdraw, line 11: Same comments here (and all the other files!) as above, but this is looking good!

docs/components/checkbox-group-field/index.md
```diff
@@ -8,11 +8,73 @@ Provides a checkbox group to be used within forms. It yields [CheckboxFields](./
...
+```
 
 ## Hint
 
-Optional. Provide a string to `@hint` to render the text into the Hint section of the fieldset.
+Optional.
+
+Use either `@hint` or `:hint`.
+
+Provide a string to `@hint` or components to `:hint` to render the text into the Hint section of the fieldset.
```
ynotdraw, line 50: Similar comments in this section as the above nit, but these are also just me being nit-picky!

docs/components/radio-field/demo/base-demo.md
```diff
@@ -9,6 +9,7 @@
   />
   <Form::Fields::Radio
     @label='Option 2'
+    @hint='Some hint'
```
ynotdraw, line 12: nit: When I wrote this initially, I didn't add a `@hint` here on purpose so that both uses could be shown (with label + with label + hint).  Would it be okay to axe this change?

docs/components/radio-field/index.md
```diff
@@ -4,7 +4,38 @@ Provides an opinionated radio element building on top of the Field component. If
...
+
+```hbs
+<Form::Fields::Radio
+  @hint='Some hint'
+  @name='options'
+  @value='option-2'
+  @onChange={{this.updateValue}}
+  @selectedValue='option-2'
+>
+    <:label>Here is a label <IconButton><Tooltip /><IconButton></:label>
```
ynotdraw, line 36: super nit: spacing here looks like 4 spaces rather than 2

packages/ember-toucan-core/src/-private/assert-block-or-argument-exists.ts
ynotdraw: yay! This looks awesome to me!

packages/ember-toucan-core/src/-private/assert-block-or-argument-exists.ts
```diff
@@ -0,0 +1,39 @@
...
+  blockExists: boolean;
+  argName: string;
+  arg?: string;
+  isRequired?: boolean;
```
ynotdraw, line 7: It may be helful to add some docs here for ourselves and future travelers:

```suggestion
  /**
   * The name of the argument that should be required.
   */
  arg?: string;
  /**
   * The name of the block or argument. Currently, the block and argument name must match.
   */
  argName: string;
  /**
   * Does the provided block exist?
   */
  blockExists: boolean;
  /**
   * Should this assertion be required, or is it optional?
   */
  isRequired?: boolean;
```

---

Another super nit - but `arg` and `argName` now that I'm looking at it _may_ be a little confusing on which is which.  I'm wondering if there's a different name _not_ including `arg` that could be used?  Would `name` work alone?  Something else?

packages/ember-toucan-core/src/-private/assert-block-or-argument-exists.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+/**
+ * @param {object} AssertBlockOrArg
+ *
+ **/
```
ynotdraw, line 13: Super nit, but I don't think we'll need this JSDoc like this here as TypeScript should handle this all for us in our editors.  It **would** be helpful though to include a comment of what this accomplishes though!

```suggestion
/**
 * This function does the thing!
 **/
```

packages/ember-toucan-core/src/components/form/fields/checkbox-group.hbs
```diff
@@ -1,20 +1,63 @@
...
+      )
+    }}
+      <field.Hint id={{field.hintId}} data-hint>
+        {{#if (has-block "hint")}}
+          {{yield to="hint"}}
+        {{else}}
+          {{@hint}}
+        {{/if}}
+      </field.Hint>
+    {{/if}}
```
ynotdraw, line 39: 👩‍🍳 this is looking great!

packages/ember-toucan-core/src/components/form/fields/file-input.ts
```diff
@@ -15,25 +18,32 @@ interface ToucanFormFileInputFieldComponentSignature {
     error?: ErrorMessage;
     files?: File[];
     hint?: string;
-    label: string;
+    label?: string;
     trigger: string;
     isDisabled?: boolean;
     multiple?: boolean;
     onChange?: (files: File[], event: FileEvent) => void;
     rootTestSelector?: string;
```
ynotdraw, line 26: As an aside / not part of this PR: it may be good to add JSDocs here like with our other components so that they show up in your editor!

packages/ember-toucan-core/src/components/form/fields/radio.hbs
```diff
@@ -17,14 +17,43 @@
...
+                {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
+              data-label
+            >
+              {{#if (has-block "label")}}
+                {{yield to="label"}}
+              {{else}}
+                {{@label}}
+              {{/if}}
+            </span>
+
```
ynotdraw, line 42: superrrr nit: looks like some extra spacing here that's not in the other files.

packages/ember-toucan-core/src/-private/assert-block-or-argument-exists.ts
```diff
@@ -0,0 +1,39 @@
...
+const assertBlockOrArgumentExists = ({
+  blockExists,
+  argName,
+  arg,
+  isRequired,
+}: AssertBlockOrArg) => {
+  if (isRequired) {
+    assert(
+      `You need either :${argName} or @${argName}`,
```
ynotdraw, line 22: question: Do you think we should put `@${argName}` first to prompt folks to use the argument rather than the named block?  Or do you think more people will use the named block version?

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -28,19 +27,33 @@ module('Integration | Component | InputField', function (hooks) {
...
+  });
+
+  test('it throws an assertion error if both `@label` and `:label` are provided', async function (assert) {
+    assert.expect(1);
+
+    setupOnerror((e: Error) => {
+      assert.strictEqual(
+        e.message,
+        'Assertion Failed: You can have :label or @label, but not both',
         'Expected an error message if @label is not provided'
```
ynotdraw, line 50: nit: It looks like this message is now incorrect.  Looks like the others do this to be a bit generic:

```suggestion
        'Expected assertion error message'
```

test-app/tests/integration/components/radio-group-field-test.gts
```diff
@@ -157,4 +169,36 @@ module('Integration | Component | RadioGroupField', function (hooks) {
...
+  });
+
+  test('it throws an assertion error if both `@label` and `:label` are provided', async function (assert) {
+    assert.expect(1);
+
+    setupOnerror((e: Error) => {
+      assert.strictEqual(
+        e.message,
+        'Assertion Failed: You can have :label or @label, but not both',
+        'Expected an error message if @label is not provided'
```
ynotdraw, line 195: nit: Similar to above, this assertion message isn't correct:

```suggestion
        'Expected assertion error message'
```

ynotdraw (commented):
This is coming together really well! I think just a few docs / nit things and then it'll be good to go! 🚀 

I also started playing around with this a bit for our next Form component on how we'll expose these named blocks over at https://github.com/CrowdStrike/ember-toucan-core/commit/3ed2f97d7180d92ba5a6472dc7dca6946b379e7c.  We should chat about this on Monday!  I'm **so excited** about this! Thanks for knocking this out!

ember-toucan-core/src/components/form/fields/checkbox-group.hbs
ynotdraw: I'm wondering if something similar to what I'm thinking for the form/core wrapper may work here.  I was hacking on this last night quite a bit and think I found something that'd work, although it's not necessarily pretty.  I'm curious if @simonihmig knows if there's a way to conditionally render named blocks when you have two-levels of components (e.g., `<Fieldset` exposes named blocks, but then another consuming component like `CheckboxGroup` wants to expose those underlying named blocks as well).

You can't conditionally render named-blocks inside of a block, **but** you _can_ conditionally render components.  So something really ugly like https://github.com/CrowdStrike/ember-toucan-core/commit/3ed2f97d7180d92ba5a6472dc7dca6946b379e7c#diff-68aa5e5601f7498e493199aca7a867ff41712f515756a155c3f7dcf1ccd45a10 may give us the best of both worlds?  Where folks can provide component arguments _or_ named blocks _or_ a combination of both.  That's what I was hacking on last night and it seems to work, although it's definitely not the best looking thing in the world.

For this case I think it'd look something like:

---

Update Fieldset to yield the new blocks like you have in this PR.

```hbs
<Form::Field as |field|>
  <fieldset
    aria-describedby="{{if @error field.errorId}} {{if @hint field.hintId}}"
    disabled={{@isDisabled}}
    ...attributes
  >
    {{! Use your new assert thing here instead, but just lazily posting the idea }}
    <legend
      class="type-md-tight text-body-and-labels block"
      data-label
    >{{@label}}{{yield to="label"}}</legend>

    {{! Use your new assert thing here instead, but just lazily posting the idea }}
    {{#if @hint}}
      <field.Hint id={{field.hintId}} data-hint>{{@hint}}{{yield to="hint"}}</field.Hint>
    {{/if}}

    <field.Control
      class="mt-1.5 flex flex-col space-y-2 rounded-sm p-1
        {{if @error 'shadow-error-outline'}}"
      data-control
    >
      {{yield}}
    </field.Control>

    {{#if @error}}
      <field.Error id={{field.errorId}} @error={{@error}} data-error />
    {{/if}}
  </fieldset>
</Form::Field>
```

---

And then in `CheckboxGroup`, use helpers to determine which mix of component arguments or named blocks are being used.  Map all properties over.

You can't do this:

```hbs
<Form::Fieldset
  @label={{@label}}
  @hint={{@hint}}
  @error={{@error}}
  @isDisabled={{@isDisabled}}
  ...attributes
>
  {{#if (has-block "label")}}
    {{yield to="label"}}
  {{/if}}

  {{#if (has-block "hint")}}
    {{yield to="hint"}}
  {{/if}}
  
  {{yield
    (hash
      CheckboxField=(component
        (ensure-safe-component this.CheckboxFieldComponent)
        name=@name
        onChange=this.handleInput
        isDisabled=@isDisabled
        selectedValues=@value
      )
    )
  }}
</Form::Fieldset>
```

But you could do something like this:

```hbs
{{#if (this.hasOnlyLabelBlock (has-block "label") (has-block "hint"))}}
  <Form::Fieldset
    @hint={{@hint}}
    @error={{@error}}
    @isDisabled={{@isDisabled}}
    ...attributes
  >
    <:label>{{yield to="label"}}</:label>
    {{yield
      (hash
        CheckboxField=(component
          (ensure-safe-component this.CheckboxFieldComponent)
          name=@name
          onChange=this.handleInput
          isDisabled=@isDisabled
          selectedValues=@value
        )
      )
    }}
  </Form::Fieldset>
{{else if (this.hasHintAndLabelBlocks (has-block "label") (has-block "hint"))}}
  <Form::Fieldset
    @hint={{@hint}}
    @error={{@error}}
    @isDisabled={{@isDisabled}}
    ...attributes
  >
    <:label>{{yield to="label"}}</:label>
    <:hint>{{yield to="hint"}}</:hint>
    {{yield
      (hash
        CheckboxField=(component
          (ensure-safe-component this.CheckboxFieldComponent)
          name=@name
          onChange=this.handleInput
          isDisabled=@isDisabled
          selectedValues=@value
        )
      )
    }}
  </Form::Fieldset>
{{else if (this.hasLabelArgAndHintBlock @label (has-block "hint"))}}
  <Form::Fieldset
    @label={{@label}}
    @hint={{@hint}}
    @error={{@error}}
    @isDisabled={{@isDisabled}}
    ...attributes
  >
    <:hint>{{yield to="hint"}}</:hint>
    {{yield
      (hash
        CheckboxField=(component
          (ensure-safe-component this.CheckboxFieldComponent)
          name=@name
          onChange=this.handleInput
          isDisabled=@isDisabled
          selectedValues=@value
        )
      )
    }}
  </Form::Fieldset>
{{else}}
  <Form::Fieldset
    @label={{@label}}
    @hint={{@hint}}
    @error={{@error}}
    @isDisabled={{@isDisabled}}
    ...attributes
  >
    {{yield
      (hash
        CheckboxField=(component
          (ensure-safe-component this.CheckboxFieldComponent)
          name=@name
          onChange=this.handleInput
          isDisabled=@isDisabled
          selectedValues=@value
        )
      )
    }}
  </Form::Fieldset>
{{/if}}
```

With your helpers here being:

```ts
  hasOnlyLabelBlock = (hasLabel: boolean, hasHint: boolean) =>
    hasLabel && !hasHint;
  hasHintAndLabelBlocks = (hasLabel: boolean, hasHint: boolean) =>
    hasLabel && hasHint;
  hasLabelArgAndHintBlock = (hasLabel: boolean, hasHint: boolean) =>
    hasLabel && hasHint;
```

nicolechung: There's also `hasHintArgAndLabelBlock` and `hasHintArgAndLabelArg` and `hasOnlyLabelArg` I think.

simonihmig: > ⚠️ You can't do this: ⚠️

What exactly can't we do?

Your example had

```hbs
{{#if (has-block "label")}}
  {{yield to="label"}}
{{/if}}
```

We can't do this?

Or is it about something like this, which is not possible? 👇

```hbs
{{#if (has-block "label")}}
  <:label>
    {{yield to="label"}}
  </:label>
{{/if}}
```

If that's the case, what is happening there? Does Ember refuse using any `<:some-block>` wrapped in `{{#if}}`?

nicolechung: Yes named blocks like `:label` can't be wrapped in conditionals. (Sorry, yesterday I was struggling to remember the exact details, but this is why!)

RFC issue: https://github.com/emberjs/rfcs/issues/735

packages/ember-toucan-core/src/components/form/fields/radio.hbs
nicolechung: The extra line space should be removed now...I notice that this isn't fixed with `pnpm format` which is a bit of a bummer.

ynotdraw (approved):
This looks good to me! Questions still around `Fieldset` though from https://github.com/CrowdStrike/ember-toucan-core/pull/110#discussion_r1155628167.  I think we can figure this out later as it seems a bit tricky due to the way named blocks work.

I also wonder if we should delete `Fieldset` as it's not being used at all with these changes and it'll be one more component to maintain that we're not sure if folks are going to use? I'd imagine most folks will use our checkbox-group and radio-group components alone rather than rolling their own, but who knows.  If we could use the `Fieldset` component for both of those cases, then I'd say let's keep it, but since we cannot due to the limitations mentioned in the those comments, I'm quite tempted to say we remove it for now (can always add it back later!).

**NOTE:** It may be wise to add a changeset with this PR as well!  Then we can cut releases more frequently 🎉.  If you do end up deleting `Fieldset` it's probably worth selecting `minor` as the upgrade (as it'll technically be a breaking change in pre1.0.0 land).  Otherwise a `patch` is probably fine since we are pre1.0.0 and it's new functionality with no breaking changes.

packages/ember-toucan-core/package.json
```diff
@@ -128,7 +128,6 @@
       "./components/form/fields/radio-group.js": "./dist/_app_/components/form/fields/radio-group.js",
       "./components/form/fields/radio.js": "./dist/_app_/components/form/fields/radio.js",
       "./components/form/fields/textarea.js": "./dist/_app_/components/form/fields/textarea.js",
-      "./components/form/fieldset.js": "./dist/_app_/components/form/fieldset.js",
```
ynotdraw, line 131: 👋

PR #115, opened 2023-03-30
Add character count like named block to input field

nicolechung:
## 🚀 Description

To be merged AFTER: https://github.com/CrowdStrike/ember-toucan-core/pull/110

Right now I have set the base branch to be `add_labels_and_error_blocks_to_fields` to make review a bit easier. Once #110 is merged I will change the base branch to `main`.

- [x]  🆕 added: `<:secondaryInformation>` block (name: TBD, suggestions?)
- [x] added: `<Form::Controls::CharacterCount>` component
  - [x] 🆕 this component is now `dumb`
  - [x] 🆕  logic for the character count is now handled by the Input Field
- [x] moved: wrapping `div` for `field.Error` into the `InputField` template
- [x] 🆕  create `field.InlineError` component
- [x] set: `flex-3` on the `field.InlineError`

## 🔬 How to Test

1. Preview link coming up

---

text-count

### Note: without the `character` block, error takes up entire width:
Screenshot 2023-03-30 at 4 37 43 PM


ember-toucan-core/src/-private/components/error.hbs
```diff
@@ -1,11 +1,7 @@
-<div
-  class="type-xs-tight text-critical mt-1.5 flex
-    {{if this.hasMoreThanOneError 'items-start' 'items-center'}}"
-  ...attributes
->
+<div ...attributes>
   {{! TODO: https://github.com/CrowdStrike/ember-toucan-core/issues/49 - Placeholder icon }}
   <svg
-    class="mr-1 h-3 w-3"
+    class="mr-1 -mt-1 inline-block h-3 w-3"
```
nicolechung, line 4: Note: added a negative margin top because the SVG did not look vertically centered. Other things I tried

- align-middle (both on this div and the parent div)

Open to suggestions on this however!

docs/components/input-field/demo/base-demo.md
```diff
@@ -16,18 +21,13 @@ import { tracked } from '@glimmer/tracking';
...
-    this.value = value;
+  handleChange(value, event) {
+    console.log({ value, event });
 
-    if (value !== 'input') {
-      this.errorMessage = 'Input must match "input"';
-      return;
-    }
-
-    this.errorMessage = null;
```
ynotdraw, line 30: Any reason to remove this block that would apply the error state? Otherwise it's a bit tricky to test the error state + character count together.

packages/ember-toucan-core/src/components/form/controls/character-count.ts
```diff
@@ -0,0 +1,48 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+
+export interface ToucanFormControlCharacterCountComponentSignature {
+  Element: HTMLElement;
```
ynotdraw, line 5: We'll want this to match the `<span` tag so:

```suggestion
  Element: HTMLSpanElement;
```

packages/ember-toucan-core/src/components/form/controls/character-count.hbs
```diff
@@ -0,0 +1,3 @@
+<span class="type-xs-tight" ...attributes>{{this.currentCount}}
+  /
+  {{@max}}</span>
```
ynotdraw, line 3: I'm wondering if there are any RTL concerns about doing this?  Is there someone we could ask for verification?

packages/ember-toucan-core/src/components/form/fields/input.hbs
```diff
@@ -35,21 +35,37 @@
...
+          @error={{@error}}
+          data-error
+        />
+      {{/if}}
+      {{#if (has-block "character")}}
+        <div class="text-body-and-labels flex-1 text-right">
+          {{yield (hash id=field.id) to="character"}}
+        </div>
+      {{/if}}
+    </div>
```
ynotdraw, line 69: I'm wondering if we can prevent another wrapping div by _only_ rendering it if there's a `has-block "character"` line?  Maybe we can leave `error.hbs` as is today and then do something like:


```hbs
{{#if (has-block "character")}}
  <div class="flex items-center">
    {{#if @error}}
      <field.Error
        class="flex-3"
        id={{field.errorId}}
        @error={{@error}}
        data-error
      />
    {{/if}}

    <div class="text-body-and-labels mt-1.5 flex-1 text-right">
      {{yield (hash id=field.id) to="character"}}
    </div>
  </div>
{{else if @error}}
  <field.Error
    {{! NOTE: Not applying flex-3 here since we don't need it }}
    id={{field.errorId}}
    @error={{@error}}
    data-error
  />
{{/if}}
```

packages/ember-toucan-core/src/components/form/controls/character-count.hbs
```diff
@@ -0,0 +1,3 @@
+<span class="type-xs-tight" ...attributes>{{this.currentCount}}
```
ynotdraw, line 1: Instead of another wrapping div over on https://github.com/CrowdStrike/ember-toucan-core/pull/115/files#diff-1f29edf2e3cdd8fbd4bbbdf9d06e2a675cbc42ffab5bc1a77b371214c86a5b46R65 for the yielded block, could we apply those styles here instead and only yield back the block?

The block:

```hbs
{{yield (hash id=field.id) to="character"}}
```

And then update this:

```hbs
<span class="type-xs-tight text-body-and-labels flex-1 text-right" ...attributes>{{this.currentCount}} / {{@max}}</span>
```

Since we are opinionated on how our character counter looks, I think it's safe to add some more layout classes here probably? I'm just wondering if we can cut down on some of the DOM elements.  If not, that's fine!

packages/ember-toucan-core/src/components/form/fields/input.hbs
```diff
@@ -35,21 +35,37 @@
         {{/if}}
       </field.Hint>
     {{/if}}
-
     <field.Control class="mt-1.5 flex">
       <Form::Controls::Input
         id={{field.id}}
         aria-describedby="{{if @hint field.hintId}} {{if @error field.errorId}}"
         aria-invalid={{if @error "true"}}
+        class="w-full"
```
ynotdraw, line 43: Do we need to add `w-full` here?  I'm just wondering - consumers may want their own widths and as you know, overriding CSS classes with Tailwind can be a pain.  Whereas if we let folks decide what they want to do instead by passing in `class` it gives them more control.


If we want this demo to be full width, we could add the following to the demo instead of here like:

```hbs
{{! demo above }}
<Form::Fields::Input class="w-full" @label='Label' />
```

docs/components/input-field/demo/base-demo.md
```diff
@@ -1,12 +1,17 @@
 ```hbs template
 <Form::Fields::Input
   @label='Label'
-  @hint='Type "input" into the field'
-  @error={{this.errorMessage}}
+  @hint='Hint text'
```
ynotdraw, line 4: Any reason to adjust the `hint` and `error` so that we can test the error state + hint state properly as part of this demo?

packages/ember-toucan-core/src/components/form/controls/character-count.ts
ynotdraw, line 1: Would it be possible to do some of the input setup behind the scenes so that users only have to do something like:

```hbs
<:character>
  <Form::Controls::CharacterCount @max={{255}}/>
</:character>
```

Maybe it could be a combination of named blocks and contextual components from `input-field`?  Where we can wire some of this up for them?  Or maybe we need modifiers?  I'm not sure, but auto-tracking the input would be dope if possible and remove some of the manual argument stuff.


Maybe in `input-field`:

```hbs
{{yield
  (hash
    CharacterCount=(component
      (ensure-safe-component this.CharacterCount)
      id=id
      currentCount=this.currentCount
    )
  )
}}
```

and then usage:

```hbs
<Form::Fields::Input
  @label='Label'
  as |input|
>
  <:character>
    <input.CharacterCount max={{255}} />
  </:character>
</Form::Fields::Input>
```

packages/ember-toucan-core/src/components/form/fields/input.ts
```diff
@@ -62,4 +67,12 @@ export default class ToucanFormInputFieldComponent extends Component<ToucanFormI
...
     return Boolean(this.args?.error);
   }
+
+  /**
+   * Used to help us determine if we should render a single error
+   * or render a ul list of errors.
+   */
+  get hasMoreThanOneError() {
+    return Array.isArray(this.args.error) && this.args.error?.length > 1;
+  }
```
ynotdraw, line 77: Keeping `error.hbs` as is as suggested at https://github.com/CrowdStrike/ember-toucan-core/pull/115/files#diff-cfd3d39e514acb00199182387cf476ea65b2cb7c105843faeac73c111413df8d means this could likely go away too!

ynotdraw (commented):
This is looking cool! Nice work!  One thing I noticed as well visually is it looks like the demo is broken with multi-line errors and not `flex`ing:

Screenshot 2023-04-12 at 10 14 10 AM


And if I add `flex`, now the error icon is offset due to the negative margin:

Screenshot 2023-04-12 at 10 19 01 AM

Some of the suggestions in this PR may resolve this though!

packages/ember-toucan-core/src/components/form/controls/character-count.ts
simonihmig: Yeah, I had the same concern. Basically we need to figure out how much freedom we want users to give in how they use this. They might come up with (crazy) ideas what to put in `<:character>` that we might not want to support at all. After all it's on us to restrict this stuff as much as we want, as this is what a design system does...

Like we could even do something as this:

```hbs
<Form::Fields::Input
  @label='Label'
  @max={{255}}
  {{! other args }}
  as |input|
/>
``` 

and render the character count component implicitly in the right place when we see that `@max` is passed!? 🤔 

Currently with...

```hbs
<:character as |data|><Form::Controls::CharacterCount
      @id={{data.id}}
      @current={{this.count}}
      @max={{255}}
    /></:character>
```

... there is a lot of things to wire up (and possible get wrong!). Might be ok if people need to really customize this deeply (again, it's on us if we want to allow for this), but the default case should be as easy as possible I think.

packages/ember-toucan-core/src/components/form/controls/character-count.ts
```diff
@@ -0,0 +1,48 @@
...
+    assert(
+      'An "@id" argument is required, the id should be from the related input',
+      args.id
+    );
+    super(owner, args);
+  }
+
+  get initialCount() {
+    const id = this.args.id;
+    const input = <HTMLInputElement>document.getElementById(id);
```
simonihmig, line 34: I would prefer to keep this component more "dumb". Can't we just make `current` a required argument? 

The normal flow of data is that we have state, and the DOM is a reflection of that state. Now we are basically pulling state (`input.value.length`) from the DOM. Might not be a huge issue here, but I would prefer to keep the other way around...

We could for example listen for `@onChange` of our `<Form::Controls::Input>`, set that as a `@tracked` property and have getter `get count()` that return the length of it. This we could then curry to the yielded character-count component, so it receives the count directly instead of assuming there must be an input somewhere and querying that.

ynotdraw: > We could for example listen for @onChange of our <Form::Controls::Input>, set that as a @tracked property and have getter get count() that return the length of it. This we could then curry to the yielded character-count component, so it receives the count directly instead of assuming there must be an input somewhere and querying that.

I like this idea!

packages/ember-toucan-core/src/components/form/controls/character-count.hbs
nicolechung: Checked with design, who checked with translation - turns out there is no issue at the moment.

packages/ember-toucan-core/src/components/form/controls/character-count.hbs
nicolechung: I removed it from this `character-count` component and added it to the demo.

docs/components/input-field/demo/base-demo.md
nicolechung: Should be reverted now!

packages/ember-toucan-core/src/-private/components/error.hbs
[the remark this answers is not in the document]
nicolechung: This should be reverted now!

packages/ember-toucan-core/src/components/form/controls/character-count.ts
nicolechung: Okay, this is gotten a bit complicated, so I updated the PR description with 🆕  which hopefully outlines what changed.

1. Spoke with design and turns out that the `character` block might have other elements in the future, so I renamed it to `secondaryInformation` for now, but happy to take suggestions for this (I was struggling with a good name for it)
2. Made `CharacterCount` a dumb component.
3. Moved the `count` logic to inside of the Input Field, but this required adding an additional `{{on "input" handler}}`. This seemed logical as this handler does separate logic from `@onChange`.

packages/ember-toucan-core/src/components/form/fields/input.ts
nicolechung: This looks like it's still needed with the `if (has-block)` condition we added.

docs/components/input-field/demo/base-demo.md
```diff
@@ -19,8 +26,9 @@ export default class extends Component {
   @tracked errorMessage;
 
   @action
-  updateValue(value, e) {
+  handleChange(value, event) {
     this.value = value;
+    console.log({ value, event });
```
simonihmig, line 31: Is this intentionally here, or forgot to remove?

packages/ember-toucan-core/src/components/form/controls/character-count.ts
simonihmig: > so I renamed it to secondaryInformation for now, but happy to take suggestions for this (I was struggling with a good name for it)

Yeah, naming is hard... maybe `secondary` is enough? But I have no strong feelings about this...

packages/ember-toucan-core/src/components/form/fields/input.ts
```diff
@@ -47,10 +52,29 @@ interface ToucanFormInputFieldComponentSignature {
...
+    secondaryInformation: [
+      {
+        CharacterCount: WithBoundArgs<typeof CharacterCount, 'current'>;
+      }
+    ];
   };
 }
 
 export default class ToucanFormInputFieldComponent extends Component<ToucanFormInputFieldComponentSignature> {
+  @tracked count = 0;
```
simonihmig, line 64:
```suggestion
  @tracked count = this.args.value?.length ?? 0;
```

I think (untested) with this you wound't need to override the constructor below, so would make it a bit simpler. A class filed initializer is basically syntactic sugar for setting the class field at the bottom of the constructor, as you do in L75, so you can do either one or the other, don't need to do both (I am not even 100% if the initializer above setting count to zero is not overriding what you do in L75!? 🤔 ), and an initializer is usually more terse and easier to read IMHO.

packages/ember-toucan-core/src/components/form/fields/input.ts
```diff
@@ -59,7 +83,29 @@ export default class ToucanFormInputFieldComponent extends Component<ToucanFormI
...
+    assert(
+      'Expected HTMLInputElement',
+      event.target instanceof HTMLInputElement
+    );
+
+    if (event.target?.value) {
+      this.count = event.target.value.length;
+    } else {
+      this.count = 0;
+    }
```
simonihmig, line 97: At the end a matter of taste, but another way to write this is:

```suggestion
    this.count = event.target?.value.length ?? 0;
```

packages/ember-toucan-core/src/components/form/controls/character-count.ts
```diff
@@ -0,0 +1,19 @@
+import Component from '@glimmer/component';
+
+export interface ToucanFormControlCharacterCountComponentSignature {
+  Element: HTMLSpanElement;
+  Args: {
+    /**
+     * The current number of characters inside the input
+     */
+    current?: number;
```
simonihmig, line 9: Are these args really optional? As you don't seem to supply a default value, so omitting them would probably render as `undefined / undefined`? I think we could keep `current` as optional, by defaulting it to zero. But `max` is probably required, as we cannot assume any default value reasonably, right?

If we allow some to be optional, then a test case for this behaviour would also be helpful!

packages/ember-toucan-core/src/components/form/field.hbs
```diff
@@ -5,6 +5,7 @@
       Hint=this.Hint
       Control=this.Control
       Error=this.Error
+      InlineError=this.InlineError
```
simonihmig, line 8: I haven't really looked into what the differences for these two error components are, but have you looked into whether it's possible to have just one, and maybe if absolutely needed pass an argument to distinguish the two modes? I am not insisting here, as again I haven't looked deeply into it and don't know the tradeoffs exactly, just thinking aloud...

docs/components/input-field/demo/base-demo.md
nicolechung: It was intentional so that folks can see what they get for `value` and `event`, but happy to remove.

packages/ember-toucan-core/src/components/form/field.hbs
ynotdraw: Would something like what I described at https://github.com/CrowdStrike/ember-toucan-core/pull/115#discussion_r1164214606 work?  Still use the existing Error component, but conditionally apply additional classes (or even use a new component argument to apply the styles we want!)?

docs/components/input-field/demo/base-demo.md
simonihmig: No, that's fine then, I just was not sure.

packages/ember-toucan-core/src/components/form/controls/character-count.ts
nicolechung: I tried making @max required, and then @current optional with something like

```javascript
// in the backing class
@tracked current = this.args.current ?? 0;

// in the constructor
this.current = this.args.current;
```

However, the @current arg was not getting updated in the test and so the test was breaking. So I went back to making `@current` required.

I would THINK that this would work, given that's how I think this works in React, but it seems like the constructor is not called in Ember everytime an argument changes??

docs/components/input-field/demo/base-demo.md
```diff
@@ -4,9 +4,16 @@
   @hint='Type "input" into the field'
   @error={{this.errorMessage}}
   @value={{this.value}}
-  @onChange={{this.updateValue}}
+  @onChange={{this.handleChange}}
+  class='w-full'
   type='text'
-/>
+>
+  <:secondaryas |secondary|><secondary.CharacterCount
```
simonihmig, line 11:
```suggestion
  <:secondary as |secondary|><secondary.CharacterCount
```

packages/ember-toucan-core/src/components/form/controls/character-count.ts
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * The max amount allowed for this input
+     */
+    max: number;
+  };
+}
+
+// eslint-disable-next-line ember/no-empty-glimmer-component-classes
+export default class ToucanFormControlCharacterCount extends Component<ToucanFormControlCharacterCountComponentSignature> {
+  @tracked current = this.args.current ?? 0;
```
simonihmig, line 22: You don't need to use `@tracked` for only providing a default value, as it is only derived state. So a getter should be sufficient. Also this is not event used in the template, it still has `@current`!

But personally I don't know if it's worth it to support an optional arg here, maybe just make it required as in the commit before?

packages/ember-toucan-core/src/components/form/fields/input.ts
```diff
@@ -47,10 +52,29 @@ export interface ToucanFormInputFieldComponentSignature {
...
+  constructor(
+    owner: unknown,
+    args: ToucanFormInputFieldComponentSignature['Args']
+  ) {
+    super(owner, args);
+
+    if (this.args.value) {
+      this.count = this.args.value.length;
+    }
+  }
```
simonihmig, line 77:
```suggestion
```

This can be removed, as L64 already does the same.

test-app/tests/integration/components/input-field-test.gts
```diff
@@ -196,4 +201,31 @@ module('Integration | Component | Fields | Input', function (hooks) {
...
+    class Context {
+      @tracked count = 0;
+
+      @action
+      handleChange(value: string) {
+        this.count = value.length;
+      }
+    }
+
+    let ctx = new Context();
```
simonihmig, line 215: Do we need this actually for this test? Seems this is never read from?

packages/ember-toucan-core/src/-private/components/inline-error.hbs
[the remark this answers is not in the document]
nicolechung: Ooouf okay reading over these comments, some things got really messed up here on my part. This file totally should have been deleted.

test-app/tests/integration/components/input-field-test.gts
nicolechung: I think it's used over on line 218 in the @onChange: https://github.com/CrowdStrike/ember-toucan-core/pull/115/files#diff-b1d74a6d42d39cd9e5f8eb62146064c2694715629721f598278442b2a9249855R218

nicolechung: For some more context, I update the `current` for the consumer (user of the component I mean) over here: https://github.com/CrowdStrike/ember-toucan-core/pull/115/files#diff-1f29edf2e3cdd8fbd4bbbdf9d06e2a675cbc42ffab5bc1a77b371214c86a5b46R70

packages/ember-toucan-core/src/components/form/controls/character-count.ts
nicolechung: Renamed to `secondary`

packages/ember-toucan-core/src/components/form/controls/character-count.ts
nicolechung: Make both required.

packages/ember-toucan-core/src/components/form/field.hbs
nicolechung: InlineError is removed.

docs/components/input-field/demo/base-demo.md
```diff
@@ -4,9 +4,16 @@
...
   @value={{this.value}}
-  @onChange={{this.updateValue}}
+  @onChange={{this.handleChange}}
+  class='w-full'
   type='text'
-/>
+>
+  <:secondary as |secondary|><secondary.CharacterCount
+      @max={{255}}
+      class="type-xs-tight"
```
ynotdraw, line 13: If this is the styling we want for character count, can it be added on the component instead of relying on consumers to always pass this in?

packages/ember-toucan-core/src/components/form/fields/input.hbs
```diff
@@ -45,10 +44,37 @@
...
+          <field.Error
+            @isInline={{true}}
+            class="flex-3"
+            id={{field.errorId}}
+            @error={{@error}}
+            data-error
+          />
+        {{/if}}
+
+        <div class="text-body-and-labels mt-1.5 flex-1 text-right">
```
ynotdraw, line 66: If we are setting the sizing on `packages/ember-toucan-core/src/components/form/controls/character-count.hbs`, so we next the text-body-and-labels class here? Should that class be set here instead so any child of the secondary block will have the font size set?

test-app/tests/integration/components/input-field-test.gts
ynotdraw: I'm wondering if we even need this Context object as well. It looks like the test has everything it needs without this, as you're only asserting the character count works as you type into it, no?

ynotdraw:
Looks like something may not be quite right with the new error styling:

Screenshot 2023-04-20 at 5 55 17 PM

simonihmig (approved):
Approved on my part, after Tony's left over concerns have been addressed! 😀

docs/components/input-field/demo/base-demo.md
ynotdraw, line 1: Could we leave the demo as-is and instead use the UI states section for the character count? It looks like we don't require `@value` or `@onChange` and with the way the character count works, it shouldn't need any backing JS/TS it looks.

packages/ember-toucan-core/src/components/form/fields/input.hbs
ynotdraw, line 66: We'll also want to remove `mt-1.5` here as it's adding additional margin top as margin is built-in to `<Error`

ynotdraw (changes requested):

PR #117, opened 2023-03-31
feat: Migrate to multi-package layout

ynotdraw:
## 🚀 Description
This moves us to a multi-package layout so that we can add more packages under `packages/*`.  The plan is to create another v2 add-on within this repo housing the glue between `ember-headless-form` and `ember-toucan-core`.

---

## 🔬 How to Test

- CI should be green

---

N/A


test-app/package.json
```diff
@@ -22,14 +22,14 @@
     "lint:js:fix": "eslint . --fix",
     "lint:types": "glint",
     "start": "ember serve",
-    "test": "npm-run-all lint \"test:!(watch)\"",
+    "test": "pnpm lint && pnpm test:ember",
```
ynotdraw, line 25: For some reason after this move, it wasn't recognizing `npm-run-all`.  Since we aren't running these in parallel, `&&` is essentially the same thing I think reading https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/npm-run-all.md.

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ concurrency:
 
 env:
   CI: true
-  dist: ember-toucan-core/dist
+  dist: packages/ember-toucan-core/dist
```
simonihmig, line 16: Idk if this setup is going to scale with multiple packages? We might consider using `turbo` here as well? (as in headless-form)

docs-app/package.json
```diff
@@ -126,7 +126,7 @@
   },
   "dependencies": {
     "@crowdstrike/ember-oss-docs": "^1.1.3",
-    "@crowdstrike/ember-toucan-core": "workspace:../ember-toucan-core",
+    "@crowdstrike/ember-toucan-core": "workspace:../packages/ember-toucan-core",
```
simonihmig, line 129: I believe `workspace:*` would be all we need here.

test-app/package.json
simonihmig: `npm-run-all` comes from some older version of the addon blueprint I think. We use `concurrently` by now, see https://github.com/CrowdStrike/ember-headless-form/blob/main/test-app/package.json#L15

simonihmig (approved):
Looks good! Some minor suggestions, but I think we can iterate on this also in future PRs (like introducing turbo)

.github/workflows/ci.yml
ynotdraw: > Idk if this setup is going to scale with multiple packages?

Yeah I think you're right here!

> We might consider using turbo here as well? (as in headless-form)

Totally!  I'll work on adding turbo right after this PR as to unblock @nicolechung with merge conflicts by me moving things around 😂 .

docs-app/package.json
ynotdraw: You're right! Just updated.  Thank you!

ynotdraw:
> npm-run-all comes from some older version of the addon blueprint I think. We use concurrently by now, see https://github.com/CrowdStrike/ember-headless-form/blob/main/test-app/package.json#L15

Went ahead and updated to use `concurrently` instead (https://github.com/CrowdStrike/ember-toucan-core/pull/117/commits/6adf5291616e544ae06823c0c17ac075f3df53c2). Thanks for the context! I removed a reference to `npm-run-all` as well at https://github.com/CrowdStrike/ember-toucan-core/pull/117/commits/cf4731066671ef611870907c32b1c13ce0d975ef

danwenzel (approved):

PR #125, opened 2023-04-10
docs: FileInput, add jsdoc

nicolechung:
## 🚀 Description

Adding missing JSDoc for the FileInput.

Related PR comment: https://github.com/CrowdStrike/ember-toucan-core/pull/110#discussion_r1159165826

- [x] added: JSDoc for File Input

## 🔬 How to Test

N / A

---

Screenshot 2023-04-10 at 5 32 35 PM

Probably looks better in VS Code.

### DIFF
--- a/packages/ember-toucan-core/src/components/form/fields/file-input.ts
+++ b/packages/ember-toucan-core/src/components/form/fields/file-input.ts
@@ -14,35 +14,36 @@ interface ToucanFormFileInputFieldComponentSignature {
   Element: HTMLInputElement;
   Args: {
     /**
-     * Provide a string to this argument, a comma separated list of file types
+     * A comma separated list of file types
      * @example: `@accept="video/*"`
      * @example: `@accept="image/png, image/jpeg"`
      */
     accept?: string;
     /**
-     * Provide a string to this argument for the delete button a11y text
+     * The delete button a11y text.
      */
     deleteLabel: string;
     /**
-     * Provide a string to this argument to render an error message and apply error styling to the Field.
+     * The error message for the file input field. Linked to the input with aria-describedby.
      */
     error?: ErrorMessage;
     /**
-     * Provide an array of File objects. This is handled automatically when a user uploads files.
+     * This array is created automatically when a user uploads files.
+     * Note that this is not a FileList object, but an array of File objects. This is for convenience as FileList does not have common array methods like filter.
      */
     files?: File[];
     /**
-     * Provide a string to this argument to render a hint message to help describe the control.
+     * Render a hint message to help describe the control. Linked to the input with aria-describedby.
      */
     hint?: string;
 
     /**
-     * Provide a string to this argument to render inside of the label tag.
+     * Renders inside the link tag.
      */
     label?: string;
 
     /**
-     * Provide a string to this argument to replace the "Select Files" text. Used for internationalization purposes.
+     * Used to replace the "Select Files" text, and used for internationalization purposes.
      */
     trigger: string;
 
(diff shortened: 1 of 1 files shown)


packages/ember-toucan-core/src/components/form/fields/file-input.ts
```diff
@@ -10,16 +10,58 @@ export type FileEvent = (Event | MouseEvent) & { target: FileTarget | null };
...
   Args: {
+    /**
+     * Provide a string to this argument, a comma separated list of file types
+     * @example: `@accept="video/*"`
+     * @example: `@accept="image/png, image/jpeg"`
+     */
     accept?: string;
+    /**
+     * Provide a string to this argument for the delete button a11y text
+     */
```
ynotdraw, line 21: Super nit, but looks like the ones below end with a `.` while these do not

packages/ember-toucan-core/src/components/form/fields/file-input.ts
```diff
@@ -10,16 +10,58 @@ export type FileEvent = (Event | MouseEvent) & { target: FileTarget | null };
...
+
+    /**
+     * Provide a boolean. Sets the multiple attribute on the file input.
+     * If not set, defaults no multiple attribute. In this case (single file upload) a file upload will REPLACE any existing file that is in the file list.
+     */
     multiple?: boolean;
+
+    /**
+      * Provide a callback to add additional functionality when files are added
+    */
```
ynotdraw, line 59: Does this also handle the case when files are removed? If so, maybe: `Provide a callback to be notified when files change.`

packages/ember-toucan-core/src/components/form/fields/file-input.ts
```diff
@@ -10,16 +10,58 @@ export type FileEvent = (Event | MouseEvent) & { target: FileTarget | null };
...
     onChange?: (files: File[], event: FileEvent) => void;
+
+    /**
+      * Provide a string. Used for an alternate named test rootTestSelector
+    */
```
ynotdraw, line 64: Same super nit comment here as above with `.`.

ynotdraw (commented):
Looks awesome! Thank you!

Could we also add some for https://github.com/CrowdStrike/ember-toucan-core/blob/main/packages/ember-toucan-core/src/components/form/controls/file-input.ts while we are here? 🙏

packages/ember-toucan-core/src/components/form/fields/file-input.ts
```diff
@@ -10,16 +10,58 @@ export type FileEvent = (Event | MouseEvent) & { target: FileTarget | null };
 interface ToucanFormFileInputFieldComponentSignature {
   Element: HTMLInputElement;
   Args: {
+    /**
+     * Provide a string to this argument, a comma separated list of file types
+     * @example: `@accept="video/*"`
+     * @example: `@accept="image/png, image/jpeg"`
```
ynotdraw, line 16: These examples are great 🚀

packages/ember-toucan-core/src/components/form/fields/file-input.ts
```diff
@@ -10,16 +10,58 @@ export type FileEvent = (Event | MouseEvent) & { target: FileTarget | null };
 interface ToucanFormFileInputFieldComponentSignature {
   Element: HTMLInputElement;
   Args: {
+    /**
+     * Provide a string to this argument, a comma separated list of file types
```
danwenzel, line 14: nit: The "Provide a string to this argument," part isn't really necessary, IMO. The type annotation lets us know it's a string, and we know these are all arguments.

```suggestion
     * A comma separated list of file types
```

(similar for the others below)

simonihmig: I agree

simonihmig (approved):
Nothing to add beyond what has already been mentioned by the other reviewers, so approving ahead of time.

PR #126, opened 2023-04-10
feat: Button Base Demo test

nicolechung:
This test passes locally, however I am running into a glint error and I am not sure how to import the demo component.

See:
Screenshot 2023-04-10 at 5 46 06 PM

See: Demo in Ember dev tools:
demo

## 🚀 Description

Another acceptable option is to put the summary of the changes in list form:

- [x] added: new test for `DocfyDemoBaseButton

-->

---

## 🔬 How to Test

N / A - first I need to fix glint error

---

N / A - first tests and lint should pass.


ynotdraw:
Would an acceptance test be more reasonable instead?

nicolechung:
> Verify no console errors?

How do you do this in a test?

docs-app/tests/acceptance/button-test.gts
```diff
@@ -0,0 +1,26 @@
...
+
+  test('visiting the button page', async function (assert) {
+
+    await visit('/docs/components/button')
+    assert.strictEqual(currentURL(), '/docs/components/button');
+
+    assert.dom('h1').hasText('Button');
+    assert.dom('.docfy-demo__example').hasTagName('div');
+
+    assert.dom('.docfy-demo__example button').hasText('Button');
```
nicolechung, line 19: It's possible to click the button here, and an `alert` will pop up, but I didn't know how to select it or hit the "okay" button so I omitted trying this.

ynotdraw:
> > Verify no console errors?
> 
> 
> 
> How do you do this in a test?

May be able to use a sinon spy

docs-app/tests/acceptance/button-test.gts
```diff
@@ -0,0 +1,40 @@
...
```
ynotdraw, line 32: Is there a data- attribute we could use that's under our control instead? Maybe not a big deal but I get anxious using class names from other libraries as selectors in tests due to being potentially brittle.

ynotdraw (approved):
This all looks good to me. I'm wondering what @simonihmig thinks here. I've seen usages of https://github.com/elwayman02/ember-sinon-qunit and looks like it may remove some of the cleanup stuff, but this looks good to me as well.

docs-app/tests/acceptance/button-test.gts
```diff
@@ -0,0 +1,40 @@
...
+    sinon.restore();
+  });
+
+  test('visiting the button page', async function (assert) {
+
+    await visit('/docs/components/button')
+    assert.strictEqual(currentURL(), '/docs/components/button');
+
+    assert.dom('h1').hasText('Button');
+    assert.dom('.docfy-demo__example').hasTagName('div');
```
simonihmig, line 30: I see little value in that assertion...

PR #128, opened 2023-04-11
Support `@isButtonGroup` mode for `<Button>`

simonihmig:
## 🚀 Description

This is adding an internal/private only(!) argument `@isButtonGroup` to `<Button>` to allow them being rendered as part of a `<ButtonGroup>` with the proper styling (rounded borders only on the outside parts). This is basically replacing some custom CSS we had previously in place to override the Tailwind-based default styling. Now the `<Button>` component "owns" all of its styling concerns, without other parts poking into its styling implementation from the outside. 

This does not feature any documentation updates, as it is _not_ supposed to be used by anything but `<ButtonGroup>` internally.

---

## 🔬 How to Test

We cannot really show the changes live, as the other part of the changes is in `@crowdstrike/ui` as part of another PR. The change here would be more obvious had we `<ButtonGroup>` already extracted into this repo.

---

image

This is a screenshot of the Toucan Guide's `ButtonGroup`, when using the other PR with these changes (using `yarn link`). See the outer sides of the buttons have rounded borders, while the inner adjacent sides don't.

### DIFF
--- a/.changeset/gold-steaks-try.md
+++ b/.changeset/gold-steaks-try.md
@@ -1,0 +1,7 @@
+---
+'@crowdstrike/ember-toucan-core': patch
+---
+
+Support `@isButtonGroup` mode for `<Button>`
+
+This is only to be used _internally_ to support buttons being used inside a `<ButtonGroup>`.
--- a/packages/ember-toucan-core/src/components/button.ts
+++ b/packages/ember-toucan-core/src/components/button.ts
@@ -19,7 +19,6 @@ const STYLES = {
     'inline-flex',
     'items-center',
     'justify-center',
-    'rounded-sm',
     'transition',
     'truncate',
     'type-md-medium',
@@ -31,6 +30,10 @@ const STYLES = {
     primary: ['interactive-primary'],
     quiet: ['font-normal', 'interactive-quiet'],
     secondary: ['interactive-normal'],
+  },
+  buttonGroup: {
+    true: ['rounded-none', 'flex-1', 'first:rounded-l-sm', 'last:rounded-r-sm'],
+    false: ['rounded-sm'],
   },
 };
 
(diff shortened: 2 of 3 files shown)


ynotdraw (commented):
Code looks great to me!  Could we add a few tests to ensure the styling in https://github.com/CrowdStrike/ember-toucan-core/blob/main/test-app/tests/integration/components/button-test.gts?

Maybe a few cases:

1) it applies rounded-small when `@isButtonGroup=false`
2) it applies the expected classes ('rounded-none', 'flex-1', 'first:rounded-l-sm', 'last:rounded-r-sm') when `@isButtonGroup=true`

simonihmig:
> Could we add a few tests to ensure the styling in https://github.com/CrowdStrike/ember-toucan-core/blob/main/test-app/tests/integration/components/button-test.gts?

Yeah, I can do that. I didn't do this right form the beginning, as we don't really have test coverage for how exactly things get styled in other cases (and I think that kind of test coverage is mostly not really useful, or better served with tools like percy). But in this case it probably makes sense, as we cannot showcase this new mode easily (again, as long as we haven't extracted `<ButtonGroup>` to this repo).

ynotdraw (approved):
🎉

nicolechung (approved):

nicolechung:
> > Maybe a few cases:
> 
> Done!
> 
> Note that I first tried to write tests using `assert.dom().hasStyle()` to assert for a particular CSS style rather than existence of CSS _classes_. But it seems Tailwind is not set up for the test-app (no styling applied at all), so that didn't work. And setting this up was a) a bit out of scope for this small PR, and b) maybe not as easy as it sounds, given that the test-app needs to run both under classic and Embroider... 🤔

I think you can do `assert.dom(DOMELEMENT).hasClass` since I see that classes are getting added: https://github.com/mainmatter/qunit-dom/blob/master/API.md#hasclass

ynotdraw:
Feel free to add a changeset to this PR and then we can probably merge it with https://github.com/CrowdStrike/ember-toucan-core/pull/127 to get a new release!

simonihmig:
> I think you can do assert.dom(DOMELEMENT).hasClass

Yeah, I have done that with the added tests.

simonihmig:
> Feel free to add a changeset to this PR and then we can probably merge it with https://github.com/CrowdStrike/ember-toucan-core/pull/127 to get a new release!

Done!

PR #129, opened 2023-04-11
Setup ember-toucan-form to yield the Textarea Field component

ynotdraw:
## 🚀 Description

This PR adds the glue between `ember-headless-form` and `ember-toucan-core` for our form solution. This PR will add `Textarea` support so that the PR scope doesn't grow substantially.

Follow up work after this PR merges:

- Add `Input` support (similar to this PR!)
- Cut a release so we can get some feedback
- Convert to using the `<template>` tag
- Checkbox support
- CheckboxGroup support
- RadioGroup support

Usage is:

```hbs template
<ToucanForm
  class='space-y-4'
  @dataMode='mutable'
  @validate={{validate-changeset}}
  as |form|
>
  <form.Textarea @label='Comment' @name='comment' />

  <Button type='submit'>Submit</Button>
</ToucanForm>
```

```js component
import {
  validatePresence,
  validateFormat,
} from 'ember-changeset-validations/validators';

export default class extends Component {
  data = {};

  validations = {
    comment: validatePresence(true),
  };

  handleSubmit(data) {
    console.log({ data });

    const { comment } = data;

    alert(`Form submitted with: ${comment}`);
  }
}
```

---

## 🔬 How to Test

- Green build
- View the installation docs at https://37bd9563.ember-toucan-core.pages.dev/docs/toucan-form
- Go to https://37bd9563.ember-toucan-core.pages.dev/docs/toucan-form/changeset-validation
  - Without typing anything into the field, click Submit
  - Verify errors are displayed
  - Type into the field
  - Click Submit
  - Verify errors go away and an alert is shown
- Do the same as above for https://37bd9563.ember-toucan-core.pages.dev/docs/toucan-form/native-validation

---

https://user-images.githubusercontent.com/8069555/231857687-06e69927-5b88-4663-9fb2-f5ab6ef63e98.mov


packages/ember-toucan-form/src/-private/textarea-field.hbs
ynotdraw, line 1: **NOTE**: My plan is to move this to the template tag stuff, but I started this in the early AM and _before_ the PR merged 🙈.  I'll get this updated maybe tomorrow, but for the time being I'm also getting a few errors I wanted to ask about.

packages/ember-toucan-form/src/-private/textarea-field.ts
```diff
@@ -0,0 +1,33 @@
...
+type ComponentArguments = BaseTextareaFieldSignature['Args'] & {
+  name: string;
+
+  // TODO: We need to type this properly
+  // What is the type when we are a component?
+  // Is this possible with TS+Glint today?
+  form: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Field: any;
+  };
```
ynotdraw, line 14: @simonihmig do you know how we can properly type this component from `ember-headless-form`?  Maybe a few issues:

1. I'm not sure how to declare this argument as a component as it's received from the parent component.  I've tried a few different syntax options like `Field: [{ value?: string }]`, etc., but Glint complains about it wanting to be an `Invokable` as it's a component.
2. The `Field` [type is exported, but under `/-private` ](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/field.ts#L27).  Should I make a PR to export this as a public interface/type so we can import it and use it here? Or should we copy/paste the types over?

I very well could be missing something silly here!

packages/ember-toucan-form/src/components/toucan-form.ts
```diff
@@ -0,0 +1,39 @@
+import Component from '@glimmer/component';
+
+import TextareaFieldComponent from '../-private/textarea-field';
+
+import type { WithBoundArgs } from '@glint/template';
+import type { HeadlessFormComponentSignature } from 'ember-headless-form/components/headless-form';
+
+// TODO: Should we export this from ember-headless-form and use it here instead? This is probably fine?
+type UserData = object;
```
ynotdraw, line 9: @simonihmig is this okay? Or do we want to export [`UserData`](https://github.com/CrowdStrike/ember-headless-form/blob/a049a6d4e275b49aa2366993aca435ccfdac9b3f/packages/ember-headless-form/src/-private/types.ts#L4)?

packages/ember-toucan-form/src/components/toucan-form.ts
```diff
@@ -0,0 +1,39 @@
...
+> {
+  Element: HTMLFormElement;
+  Args: HeadlessFormArguments<DATA, SUBMISSION_VALUE>;
+  Blocks: {
+    default: [
+      {
+        Textarea?: WithBoundArgs<typeof TextareaFieldComponent, 'form'>;
+        // TODO: Need to figure out a type for this
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        Field: any;
```
ynotdraw, line 28: @simonihmig kind of similar to above - we want a type for this, so maybe exposing a type from headless-form we can use, define it here ourselves, or something else?

packages/ember-toucan-form/src/-private/textarea-field.hbs
```diff
@@ -0,0 +1,13 @@
...
+    @label={{@label}}
+    @hint={{@hint}}
+    @error={{this.mapErrors field.rawErrors}}
+    @value={{field.value}}
+    @onChange={{field.setValue}}
+    @isDisabled={{@isDisabled}}
+    @rootTestSelector={{@rootTestSelector}}
+    name={{@name}}
+    ...attributes
+  />
```
ynotdraw, line 12: I'm getting the following lint error:

```bash
@crowdstrike/ember-toucan-form:lint: [lint:types] src/-private/textarea-field.hbs:2:3 - error TS7053: Unknown name 'Form::Fields::Textarea'. If this isn't a typo, you may be missing a registry entry for this value; see the Template Registry page in the Glint documentation for more details.
@crowdstrike/ember-toucan-form:lint: [lint:types]   Element implicitly has an 'any' type because expression of type '"Form::Fields::Textarea"' can't be used to index type 'Globals'.
@crowdstrike/ember-toucan-form:lint: [lint:types]     Property 'Form::Fields::Textarea' does not exist on type 'Globals'.
@crowdstrike/ember-toucan-form:lint: [lint:types] 
@crowdstrike/ember-toucan-form:lint: [lint:types]   2   <Form::Fields::Textarea
@crowdstrike/ember-toucan-form:lint: [lint:types]       ~~~~~~~~~~~~~~~~~~~~~~~
@crowdstrike/ember-toucan-form:lint: [lint:types]   3     @label={{@label}}
@crowdstrike/ember-toucan-form:lint: [lint:types]     ~~~~~~~~~~~~~~~~~~~~~
@crowdstrike/ember-toucan-form:lint: [lint:types] ... 
@crowdstrike/ember-toucan-form:lint: [lint:types]  11     ...attributes
@crowdstrike/ember-toucan-form:lint: [lint:types]     ~~~~~~~~~~~~~~~~~
@crowdstrike/ember-toucan-form:lint: [lint:types]  12   />
@crowdstrike/ember-toucan-form:lint: [lint:types]     ~~~~
```

Which is weird because I'm extending the unpublished registry with Toucan Core's registry at https://github.com/CrowdStrike/ember-toucan-core/pull/129/files#diff-84c037d42edbe42211468688395a6b1a14a5bbaee06c5b48b6125cfe93c48255R31

---

I'm also getting this:

```bash
[lint:types] ../../node_modules/.pnpm/ember-headless-form@1.0.0-beta.1_@glimmer+componen_ym5tcvhcc62nd3g4tiyqwpfw6i/node_modules/ember-headless-form/dist/components/headless-form.d.ts:135:16 - error TS2307: Cannot find module '@ember/modifier' or its corresponding type declarations.
[lint:types] 
[lint:types] 135     on: import("@ember/modifier").OnModifier;
[lint:types]                    ~~~~~~~~~~~~~~~~~
[lint:types] 
```

Which I added `ember-modifier` as a dependency for as well.  Any ideas?

packages/ember-toucan-form/package.json
```diff
@@ -29,19 +29,23 @@
     "test": "echo 'A v2 addon does not have tests, run tests in test-app'",
     "prepack": "rollup --config",
     "format:check": "prettier -c .",
-    "format": "prettier -w ."
+    "format": "prettier -w .",
+    "_syncPnpm": "pnpm sync-pnpm"
   },
   "peerDependencies": {
+    "@crowdstrike/ember-toucan-core": "workspace:*",
```
ynotdraw, line 36: Since this is a peer dependency, I feel like this should be a version specified for consumers rather than what I have added here.  So if someone installs `@crowdstrike/ember-toucan-form`, we also want them to be on some version of `@crowdstrike/ember-toucan-core` as well as that'll be used.  With this syntax though, I'm not sure how that'll quite work since this is an add-on consuming yet another add-on.

I feel like this should be the following instead:

```suggestion
    "@crowdstrike/ember-toucan-core": "^0.1.0",
```

But then how about the `devDependency` below?  Should that remain `workspace:*` with the pnpm dependency syncing?  I haven't done something like this in a monorepo before so I'm a bit out of my comfort zone.

simonihmig: > I feel like this should be the following instead:

Right! `workspace:*` is a pnpm specific thing AFAIK, so we cannot expose this to consumers. It should be what you suggested. Same [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/changeset/package.json#L32).

> But then how about the devDependency below? Should that remain workspace:* with the pnpm dependency syncing?

I used the same specifier [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/changeset/package.json#L43). Pnpm will still use the code from the repo rather than pulling the actually published version from npm, as long as the versions match. At least that's my understanding. But `workspace:*` will _guarantee_ that you will always get the code from the repo, and _never_ pull from npm. Which is right I think. So we could try that for our devDependencies, and see how it goes...

packages/ember-toucan-form/src/-private/textarea-field.ts
simonihmig: I think this has to match what the `<HeadlessForm>` component yields [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/components/headless-form.ts#L89-L128) - and we shouldn't try to redefine it, instead reuse the types that headless-form already provides. However I am not sure how to only get the yielded block, when the only public type that we can import is that of the [`HeadlessForm` component](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/index.ts#L10). A Glint helper like `GetBlockParams(HeadlessForm, 'default')[0]` would be nice, but it doesn't seem to exist...

Maybe we can build such a helper, but I cannot do this off the top of my head, without actually working on the code 😅. I can give it a shot tomorrow maybe if you want? Not sure if my TS skills will be sufficient though...

packages/ember-toucan-form/src/components/toucan-form.ts
simonihmig: Yeah, it think it's ok. We could export it, but nothing that should block us here...

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
```diff
@@ -0,0 +1,102 @@
...
+    await render(<template>
+      <ToucanForm>
+        <div data-custom-content />
+      </ToucanForm>
+    </template>);
+
+    assert.dom('[data-custom-content]').exists();
+  });
+
+  test('it yields a Field from ember-headless-form', async function (assert) {
```
simonihmig, line 29: I am not opposed to doing this, but I wonder what exactly is the use case you have in mind here? Supporting custom controls that are not provided by `<ToucanForm>`?

packages/ember-toucan-form/src/components/toucan-form.ts
simonihmig: Yeah, it's basically solved when the other problem is solved: when we have the type for `form`, then this is as simple as `form['Field']`...

packages/ember-toucan-form/src/components/toucan-form.hbs
```diff
@@ -0,0 +1,20 @@
+<HeadlessForm
+  @data={{@data}}
+  @dataMode={{@dataMode}}
+  @validateOn={{@validateOn}}
+  @revalidateOn={{@revalidateOn}}
+  @validate={{@validate}}
+  @onSubmit={{@onSubmit}}
+  @onInvalid={{@onInvalid}}
```
simonihmig, line 8: We can expose everything as we do here. But we _could_ also have opinions and not let users define everything on their own. Like for example we could define `@validateOn="focusout" and @revalidateOn="input"`, to make all forms behave consistently across our design system.

It's a design choice basically. Maybe worth discussing in one of the next jam sessions?

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
ynotdraw: Yup precisely! I envision folks needing/wanting to build their own custom form controls and by yielding back out `Field` it'll help them to do so.  I could be pre-optimizing though 😉.  What do you think?  If we find this is a real need, we can always add it back later.

packages/ember-toucan-form/src/components/toucan-form.hbs
ynotdraw: For sure! I'll bring this up today!  Excellent call.

packages/ember-toucan-form/src/-private/textarea-field.ts
ynotdraw: Oh that'd be pretty cool if we could make that work!  If you have some time tomorrow, that'd be great.  I'll also be on in the AM if you'd like to pair and see if we can arrive at something.  Maybe with our powers combined we can figure this out! 🎉

packages/ember-toucan-form/package.json
ynotdraw: Nice, thank you!  I'll go ahead and get this updated.  I'll give it a shot in devDependencies with `workspace:*` and see what happens

docs/toucan-form/changeset-validation/demo/base-demo.md
```diff
@@ -0,0 +1,39 @@
+```hbs template
+<ToucanForm
+  class='space-y-4'
+  @data={{changeset this.data this.validations}}
+  @dataMode='mutable'
+  @onSubmit={{this.handleSubmit}}
+  @validate={{validate-changeset}}
```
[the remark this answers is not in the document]
nicolechung: It looks like it's coming from `ember-headless-form` still but if so that should be mentioned in the docs...did I mention this already?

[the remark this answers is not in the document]
ynotdraw: Yup `validate-changeset` comes from `ember-headless-form`.  I'll get something added about this - thanks!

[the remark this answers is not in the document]
simonihmig: That's from the optional adapter package `@ember-headless-form/changeset`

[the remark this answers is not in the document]
ynotdraw: Just pushed some new docs!

packages/ember-toucan-form/src/components/toucan-form.hbs
ynotdraw: Went ahead and added a different `validateOn` default, but we want to test this out in some real apps and get feedback before adjusting `revalidateOn`'s default

docs/toucan-form/changeset-validation/index.md
```diff
@@ -0,0 +1,48 @@
...
+
+const data: MyFormData = {};
+```
+
+```hbs
+<ToucanForm @data={{data}} as |form|>
+  <form.Input @label='First name' @name='firstName' />
+  <form.Input @label='Last name' @name='lastName' />
+</ToucanForm>
+```
```
ynotdraw, line 48: You'll notice this section is copied across here and the native-validation docs as well.  I was thinking about putting these in the root https://f37cfe54.ember-toucan-core.pages.dev/docs/toucan-form, but I personally tend to go to the code examples with documentation anyway.  I'd be afraid if we put this information at https://f37cfe54.ember-toucan-core.pages.dev/docs/toucan-form, people just won't read it, as they _maybe_ only care about a) looking at code b) either the changeset or native demos anyway.

Any thoughts?  I don't have strong opinions here, more just wanted to share what I was thinking.

docs/installation.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Getting Started
+order: 1
+---
```
ynotdraw, line 4: This change moves the installation step as the first one in the nav.  Otherwise the Toucan Form stuff would be at the top

docs/toucan-form/changeset-validation/index.md
```diff
@@ -0,0 +1,48 @@
...
+pnpm add ember-headless-form-changeset
+# or
+yarn add ember-headless-form-changeset
+# or
+npm install ember-headless-form-changeset
+```
+
+## Default initial validation
+
+By default, Toucan Form will set the `@validateOn` argument to `focusout`; however, this can be overriden by providing your own `@validateOn` component argument [supported by ember-headless-form](https://ember-headless-form.pages.dev/docs/validation/timing#validateon).
```
nicolechung, line 24:
```suggestion
By default, Toucan Form will set the `@validateOn` argument to `focusout` (meaning: validation will occur when the user tabs away)
```

docs/toucan-form/changeset-validation/index.md
```diff
@@ -0,0 +1,48 @@
...
+order: 2
+---
+
+# Ember Changeset validation
+
+This demo shows how to implement [ember-changeset](https://github.com/poteto/ember-changeset) validation with ember-toucan-form, powered by [ember-headless-form](https://ember-headless-form.pages.dev/docs/validation/ember-changeset).
+
+## Install the adapter package
+
+Before using ember-changeset validations with Toucan Form, you'll need to install it as a dependency.
```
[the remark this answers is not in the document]
ynotdraw: Doh, I forgot about this.  Thanks!  Adding now.

nicolechung (approved):

docs/toucan-form/changeset-validation/index.md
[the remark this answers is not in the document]
ynotdraw: Just added in https://github.com/CrowdStrike/ember-toucan-core/pull/129/commits/8f71ed12ccca1307ac201316c4ae1bca39f9208c - give it a look and let me know if you have any suggestions!

ynotdraw:
For some reason the `Run marocchino/sticky-pull-request-comment@v2` step is failing.  I'm going to file an issue

docs/toucan-form/changeset-validation/index.md
```diff
@@ -6,3 +6,48 @@ order: 2
...
+
+By default, Toucan Form will set the `@validateOn` argument to `focusout` (meaning: validation will occur when the user tabs away); however, this can be overriden by providing your own `@validateOn` component argument [supported by ember-headless-form](https://ember-headless-form.pages.dev/docs/validation/timing#validateon).
+
+```hbs
+<ToucanForm @validateOn='input' />
+```
+
+## Form data and TypeScript
+
+The provided form `@data` argument must be properly typed if you are using TypeScript. Follow the ember-headless-form [typing of form data docs](https://ember-headless-form.pages.dev/docs/typescript#typing-of-form-data) for more information. Essentially your `@data` argument type must properly match the form fields used in your templates.
```
simonihmig, line 37: :+1:

packages/ember-toucan-form/package.json
```diff
@@ -29,19 +29,23 @@
...
-    "format": "prettier -w ."
+    "format": "prettier -w .",
+    "_syncPnpm": "pnpm sync-pnpm"
   },
   "peerDependencies": {
+    "@crowdstrike/ember-toucan-core": "^0.1.0",
     "@crowdstrike/ember-toucan-styles": "^2.0.1",
     "@ember/test-helpers": "^2.8.1",
     "@glimmer/tracking": "^1.1.2",
+    "ember-headless-form": "1.0.0-beta.1",
```
simonihmig, line 40: 1. we should probably get a 1.0.0 release out here as well.
2. we should allow for future versions:
```suggestion
    "ember-headless-form": "^1.0.0-beta.1",
```

simonihmig (approved):
Just one little package.json detail to fix, otherwise looks good! Exciting! 🚀

packages/ember-toucan-form/package.json
ynotdraw: Perfect, just updated! 🎉

PR #130, opened 2023-04-11
changeset: exit prerelease mode

simonihmig:
## 🚀 Description

This is the result of running `changeset pre exit`, to switch to doing regular `latest` 0.1.x releases instead of `0.1.0-beta.x` releases, see https://github.com/CrowdStrike/ember-toucan-core/pull/127#discussion_r1163227412

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,5 +1,5 @@
 {
-  "mode": "pre",
+  "mode": "exit",
   "tag": "beta",
   "initialVersions": {
     "docs-app": "0.0.0",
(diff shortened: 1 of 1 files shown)


.changeset/pre.json
```diff
@@ -1,11 +1,13 @@
 {
-  "mode": "pre",
+  "mode": "exit",
   "tag": "beta",
```
ynotdraw, line 3: I'm on mobile, but I can't remember - do we still need this?

nicolechung (approved):

ynotdraw (approved):

.changeset/pre.json
simonihmig: Not sure how this will actually work, been following https://github.com/changesets/changesets/blob/main/docs/prereleases.md

Maybe this will get changed later with the release PR of the changeset bot!? 🤔

ynotdraw: Yeah! Let's find out! Can always adjust if it's not as we expect

PR #133, opened 2023-04-13
Add better typing for ember-toucan-form

simonihmig:
This is improving the typing of #129, see https://github.com/CrowdStrike/ember-toucan-core/pull/129#discussion_r1163240104

### DIFF
--- a/packages/ember-toucan-form/src/-private/textarea-field.hbs
+++ b/packages/ember-toucan-form/src/-private/textarea-field.hbs
@@ -3,7 +3,8 @@
     @label={{@label}}
     @hint={{@hint}}
     @error={{this.mapErrors field.rawErrors}}
-    @value={{field.value}}
+    @value={{this.assertString field.value}}
+    {{! @glint-expect-error }}
     @onChange={{field.setValue}}
     @isDisabled={{@isDisabled}}
     @rootTestSelector={{@rootTestSelector}}
--- a/packages/ember-toucan-form/src/-private/textarea-field.ts
+++ b/packages/ember-toucan-form/src/-private/textarea-field.ts
@@ -1,33 +1,57 @@
 import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
 
+import type { HeadlessFormBlock, UserData } from './types';
 import type { ToucanFormTextareaFieldComponentSignature as BaseTextareaFieldSignature } from '@crowdstrike/ember-toucan-core/components/form/fields/textarea';
+import type { FormData, FormKey, ValidationError } from 'ember-headless-form';
 
-type ComponentArguments = BaseTextareaFieldSignature['Args'] & {
-  name: string;
+type ComponentArguments<
+  DATA extends UserData,
+  KEY extends FormKey<FormData<DATA>> = FormKey<FormData<DATA>>
+> = Omit<BaseTextareaFieldSignature['Args'], 'error' | 'value' | 'onChange'> & {
+  /**
+   * The name of your field, which must match a property of the `@data` passed to the form
+   */
+  name: KEY;
 
-  // TODO: We need to type this properly
-  // What is the type when we are a component?
-  // Is this possible with TS+Glint today?
-  form: {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    Field: any;
-  };
+  /*
+   * @internal
+   */
+  form: HeadlessFormBlock<DATA>;
 };
 
-export interface ToucanFormTextareaFieldComponentSignature {
+export interface ToucanFormTextareaFieldComponentSignature<
+  DATA extends UserData,
+  KEY extends FormKey<FormData<DATA>> = FormKey<FormData<DATA>>
+> {
   Element: HTMLTextAreaElement;
-  Args: ComponentArguments;
+  Args: ComponentArguments<DATA, KEY>;
   Blocks: {
     default: [];
   };
 }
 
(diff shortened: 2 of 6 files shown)


packages/ember-toucan-form/src/-private/textarea-field.hbs
```diff
@@ -3,7 +3,8 @@
     @label={{@label}}
     @hint={{@hint}}
     @error={{this.mapErrors field.rawErrors}}
-    @value={{field.value}}
+    @value={{this.assertString field.value}}
```
simonihmig, line 6: This is similar to https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/field.ts#L256-L265. `field.value` could be anything, but `@value` only accepts strings. 

Had we types `USERDATA` as `Record<string, string>`, then we wouldn't have this issue. But it's just a generic `object`, as we want to allow arbitrary values. Like a checkbox group would be represented by an array...

packages/ember-toucan-form/src/-private/textarea-field.ts
```diff
@@ -1,33 +1,57 @@
...
+import type { HeadlessFormBlock, UserData } from './types';
 import type { ToucanFormTextareaFieldComponentSignature as BaseTextareaFieldSignature } from '@crowdstrike/ember-toucan-core/components/form/fields/textarea';
+import type { FormData, FormKey, ValidationError } from 'ember-headless-form';
 
-type ComponentArguments = BaseTextareaFieldSignature['Args'] & {
-  name: string;
+type ComponentArguments<
+  DATA extends UserData,
+  KEY extends FormKey<FormData<DATA>> = FormKey<FormData<DATA>>
+> = Omit<BaseTextareaFieldSignature['Args'], 'error' | 'value' | 'onChange'> & {
```
simonihmig, line 11: We are setting `'error' | 'value' | 'onChange'` ourselves, so we shouldn't let the user supply them. Only what we are actually passing through should be valid. Instead of `Omit<>` we could also use `Pick<>` to whitelist the passed through arguments, maybe that's better?

packages/ember-toucan-form/src/-private/types.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { ComponentSignatureBlocks } from '@glint/template/-private/signature';
+import type { HeadlessForm } from 'ember-headless-form';
+
+export type UserData = object;
+
+// Glint utilities
+
+export type SignatureOf<C> = C extends typeof Component<infer S> ? S : never;
+
+export type GetBlockParams<
```
simonihmig, line 11: This helper type is basically the key to infer what our `form` is (yielded from headless-form), by just passing the type of `HeadlessForm` (which is a public type), see L21 below.

packages/ember-toucan-form/src/components/toucan-form.ts
```diff
@@ -34,7 +31,7 @@ export default class ToucanFormComponent<
   DATA extends UserData,
   SUBMISSION_VALUE
 > extends Component<ToucanFormComponentSignature<DATA, SUBMISSION_VALUE>> {
-  TextareaFieldComponent = TextareaFieldComponent;
+  TextareaFieldComponent = TextareaFieldComponent<DATA>;
```
simonihmig, line 34: Lost most time on this one. Same back then when working on headless-form, but forgot about this... 🙈😅

This is important to propagate the type of `DATA` down to the yielded components, otherwise TS/Glint would think that the `DATA` of the form could be of a different type than the `DATA` of e.g. the TextareaField...

simonihmig:
CI is failing, at least some typing issues in the demos, but this is not targeting `main` but the existing WIP PR, so maybe we can just ignore this and fix later?

packages/ember-toucan-form/src/-private/textarea-field.ts
```diff
@@ -1,33 +1,57 @@
...
-  // What is the type when we are a component?
-  // Is this possible with TS+Glint today?
-  form: {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    Field: any;
-  };
+  /*
+   * @internal
+   */
+  form: HeadlessFormBlock<DATA>;
```
ynotdraw, line 20: 🤯 this is awesome!

ynotdraw (approved):
🧙 thanks so much! This is excellent 🚀

ynotdraw:
> CI is failing, at least some typing issues in the demos, but this is not targeting `main` but the existing WIP PR, so maybe we can just ignore this and fix later?

I'm a bit curious about this one:

```bash
test-app:lint: [lint:types] tests/integration/components/toucan-form/toucan-form-test.gts:32:22 - error TS2322: Type 'string' is not assignable to type 'never'.
test-app:lint: [lint:types] 
test-app:lint: [lint:types] 32         <form.Field @name="field" as |field|>
```

It's acting like it doesn't like the `@name` argument?

simonihmig:
> Do we need to provide @data to satisfy it?

Yeah, exactly! Just did that. We might also need some guides here similar to https://ember-headless-form.pages.dev/docs/typescript#typing-of-form-data, or at least point to those...

ynotdraw:
> We might also need some guides here similar to https://ember-headless-form.pages.dev/docs/typescript#typing-of-form-data, or at least point to those...

Good point! I'll get some docs added over on https://github.com/CrowdStrike/ember-toucan-core/pull/129 about that once this merges 🎉

nicolechung (approved):

PR #152, opened 2023-04-26
docs: Fix changeset demo with helper

ynotdraw:
## 🚀 Description

This updates the `validate-changeset` helper to work as expected

---

## 🔬 How to Test

- Visit https://b7e12591.ember-toucan-core.pages.dev/docs/toucan-form/changeset-validation
  - Click the Submit button
  - Verify errors are displayed

In comparison, if you go to https://ember-toucan-core.pages.dev/docs/toucan-form/changeset-validation and do the same, you'll see the form submits.

---

N/A

### DIFF
--- a/docs/toucan-form/changeset-validation/demo/base-demo.md
+++ b/docs/toucan-form/changeset-validation/demo/base-demo.md
@@ -4,7 +4,7 @@
   @data={{changeset this.data this.validations}}
   @dataMode='mutable'
   @onSubmit={{this.handleSubmit}}
-  @validate={{validate-changeset}}
+  @validate={{(validate-changeset)}}
   as |form|
 >
   <form.Input @label='First name' @name='firstName' />


docs/toucan-form/changeset-validation/demo/base-demo.md
```diff
@@ -4,7 +4,7 @@
   @data={{changeset this.data this.validations}}
   @dataMode='mutable'
   @onSubmit={{this.handleSubmit}}
-  @validate={{validate-changeset}}
+  @validate={{(validate-changeset)}}
```
nicolechung, line 7: Just for my own knowledge, what's the `()` add? Calling the `validate-changeset`?

nicolechung (approved):

docs/toucan-form/changeset-validation/demo/base-demo.md
simonihmig: Exactly!

It's a bit weird tbh. For some earlier versions of ember this syntax didn't exist IIRC. It was added so you can call a helper without any arguments. Compare this to e.g. `{{validate-changeset foo=bar}}`. This is also calling, because there is no way to pass arguments to a function/helper other than calling it. And `{{validate-changeset}}` is used to pass things as a value, so in this case passing the helper function, instead of calling it. So Ember needed an explicit way to call a helper, when no args are expected, and this is the syntax for that!

simonihmig (approved):

PR #168, opened 2023-06-06
chore: remove unused ember-cached-decorator-polyfill

simonihmig:
We don't need this, as we are already on an Ember version that supports `@cached` natively (>= 4.1).

This will also fix running into https://github.com/CrowdStrike/ember-headless-form/issues/146 for the docs-app.


ynotdraw (approved):

simonihmig:
Hm, publishing the docs failed to cloudflare failed with an auth error. I though this would only happen for external contributors (as secrets are not available in workflow runs)... 🤔

PR #177, opened 2023-06-13
Add FileInput support to toucan-form

ynotdraw:
## 🚀 Description

This PR adds `form.FileInput` support to `toucan-form`.

```hbs
<ToucanForm
  @validate={{(validate-changeset)}}
  as |form|
>
  <form.FileInput
    @label='Files to attach'
    @trigger='Select files'
    @deleteLabel='Delete'
    @name='files'
  />

  <Button type='submit'>Submit</Button>
</ToucanForm>
```

---

## 🔬 How to Test

- Green build
- Go to https://e6846db9.ember-toucan-core.pages.dev/docs/toucan-form/changeset-validation
  - Click submit button
  - Verify error is displayed under FileInput component
  - Click Select files button
  - Add a file
  - Verify error goes away
  - Complete the rest of the form
  - Submit it
  - Verify in the alert `files` has content

---

https://github.com/CrowdStrike/ember-toucan-core/assets/8069555/200e10f4-fbec-4c9c-95d5-7e09b6584994

### DIFF
--- a/packages/ember-toucan-core/src/components/form/controls/file-input.hbs
+++ b/packages/ember-toucan-core/src/components/form/controls/file-input.hbs
@@ -1,4 +1,5 @@
 <input
+  accept={{this.accept}}
   class="text-titles-and-attributes focus:outline-none absolute m-1 h-6 w-0 cursor-pointer bg-transparent p-1
     {{if @isDisabled 'text-disabled' 'text-titles-and-attributes'}}"
   disabled={{@isDisabled}}
--- a/packages/ember-toucan-form/src/-private/file-input-field.hbs
+++ b/packages/ember-toucan-form/src/-private/file-input-field.hbs
@@ -20,17 +20,17 @@
   {{#if (this.hasOnlyLabelBlock (has-block 'label') (has-block 'hint'))}}
     <Form::Fields::FileInput
       @accept={{@accept}}
-      @hint={{@hint}}
+      @deleteLabel={{@deleteLabel}}
       @error={{this.mapErrors field.rawErrors}}
       @files={{this.assertArrayOfFiles field.value}}
-      {{! @glint-expect-error }}
-      @onChange={{field.setValue}}
+      @hint={{@hint}}
       @isDisabled={{@isDisabled}}
       @isReadOnly={{@isReadOnly}}
-      @rootTestSelector={{@rootTestSelector}}
       @multiple={{@multiple}}
+      {{! @glint-expect-error }}
+      @onChange={{field.setValue}}
+      @rootTestSelector={{@rootTestSelector}}
       @trigger={{@trigger}}
-      @deleteLabel={{@deleteLabel}}
       name={{@name}}
       ...attributes
     >
@@ -40,16 +40,16 @@
   }}
     <Form::Fields::FileInput
       @accept={{@accept}}
+      @deleteLabel={{@deleteLabel}}
       @error={{this.mapErrors field.rawErrors}}
       @files={{this.assertArrayOfFiles field.value}}
-      {{! @glint-expect-error }}
-      @onChange={{field.setValue}}
       @isDisabled={{@isDisabled}}
       @isReadOnly={{@isReadOnly}}
-      @rootTestSelector={{@rootTestSelector}}
       @multiple={{@multiple}}
+      {{! @glint-expect-error }}
+      @onChange={{field.setValue}}
+      @rootTestSelector={{@rootTestSelector}}
       @trigger={{@trigger}}
-      @deleteLabel={{@deleteLabel}}
       name={{@name}}
       ...attributes
     >
@@ -59,17 +59,17 @@
   {{else if (this.hasLabelArgAndHintBlock @label (has-block 'hint'))}}
     <Form::Fields::FileInput
       @accept={{@accept}}
-      @label={{@label}}
+      @deleteLabel={{@deleteLabel}}
       @error={{this.mapErrors field.rawErrors}}
       @files={{this.assertArrayOfFiles field.value}}
-      {{! @glint-expect-error }}
-      @onChange={{field.setValue}}
       @isDisabled={{@isDisabled}}
       @isReadOnly={{@isReadOnly}}
-      @rootTestSelector={{@rootTestSelector}}
+      @label={{@label}}
       @multiple={{@multiple}}
+      {{! @glint-expect-error }}
+      @onChange={{field.setValue}}
+      @rootTestSelector={{@rootTestSelector}}
       @trigger={{@trigger}}
-      @deleteLabel={{@deleteLabel}}
       name={{@name}}
       ...attributes
     >
@@ -79,18 +79,18 @@
     {{! Argument-only case }}
     <Form::Fields::FileInput
       @accept={{@accept}}
-      @label={{@label}}
-      @hint={{@hint}}
+      @deleteLabel={{@deleteLabel}}
       @error={{this.mapErrors field.rawErrors}}
       @files={{this.assertArrayOfFiles field.value}}
-      {{! @glint-expect-error }}
-      @onChange={{field.setValue}}
+      @hint={{@hint}}
       @isDisabled={{@isDisabled}}
       @isReadOnly={{@isReadOnly}}
-      @rootTestSelector={{@rootTestSelector}}
+      @label={{@label}}
       @multiple={{@multiple}}
+      {{! @glint-expect-error }}
+      @onChange={{field.setValue}}
+      @rootTestSelector={{@rootTestSelector}}
       @trigger={{@trigger}}
-      @deleteLabel={{@deleteLabel}}
       name={{@name}}
       ...attributes
     />
--- a/packages/ember-toucan-form/src/-private/file-input-field.hbs
+++ b/packages/ember-toucan-form/src/-private/file-input-field.hbs
@@ -1,0 +1,98 @@
+{{!
+  Regarding Conditionals
+
+  This looks really messy, but Form::Fields::FileInput exposes named blocks; HOWEVER,
+  we cannot conditionally render named blocks due to https://github.com/emberjs/rfcs/issues/735.
+
+  We *can* conditionally render components though, based on the blocks and argument combinations
+  users provide us.  This is very brittle, but until https://github.com/emberjs/rfcs/issues/735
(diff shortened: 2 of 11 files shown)


test-app/tests/integration/components/toucan-form/toucan-form-test.gts
```diff
@@ -119,12 +132,30 @@ module('Integration | Component | ToucanForm', function (hooks) {
...
+      // submitted values.
+      assert.ok(files, 'Expected files to be included');
+      assert.strictEqual(
+        files?.length,
+        1,
+        'Expected only a single file to be added'
+      );
+      assert.strictEqual(files?.[0]?.name, 'file.txt');
+      assert.strictEqual(files?.[0]?.type, 'text/plain');
+      assert.strictEqual(files?.[0]?.size, 19);
```
ynotdraw, line 156: Super open to any suggestions here, but figured this is probably safe enough.  I just didn't want to get into the business of comparing `lastModified` and `lastModifiedDate`.

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
```diff
@@ -119,12 +132,30 @@ module('Integration | Component | ToucanForm', function (hooks) {
...
+      // to properties on the File object prototype like `arrayBuffer, `lastModifiedDate`,
+      // and `lastModified`.  Due to that, we will assert `files` separately from the other
+      // submitted values.
+      assert.ok(files, 'Expected files to be included');
+      assert.strictEqual(
+        files?.length,
+        1,
+        'Expected only a single file to be added'
+      );
+      assert.strictEqual(files?.[0]?.name, 'file.txt');
```
clintcs, line 154: I'm not suggesting a change. Just opining. 

I've gotten into the habit of using `Array.prototype.at` for non-tuple array-index access. It produces the same result, of course, but `Array.prototype.at`'s return type includes `undefined`, which forces me to assert against `undefined` instead of possibly allowing a runtime error to propagate. 

Unicorn [has a rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-at.md) for it. Unfortunately, I can't find a one-off plugin with the rule.

packages/ember-toucan-form/src/-private/file-input-field.hbs
```diff
@@ -0,0 +1,98 @@
+{{!
+  Regarding Conditionals
+
+  This looks really messy, but Form::Fields::FileInput exposes named blocks; HOWEVER,
+  we cannot conditionally render named blocks due to https://github.com/emberjs/rfcs/issues/735.
+
+  We *can* conditionally render components though, based on the blocks and argument combinations
```
clintcs, line 7: 😔

clintcs (approved):

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
clintcs: Though `FileList` is array-like and so doesn't support `.at`. It does support `.item`.

packages/ember-toucan-form/src/-private/file-input-field.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  @action
+  assertArrayOfFiles(value: unknown): File[] | undefined {
+    assert(
+      `Only File[] values are expected for ${String(
+        this.args.name
+      )}, but you passed ${typeof value}`,
+      typeof value === 'undefined' ||
+        Array.isArray(value) ||
+        (value as File[]).forEach((file) => file instanceof File)
```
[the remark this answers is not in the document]
ynotdraw: omg, excellent catch 🙈.  One thing that was adjusted with file-input is instead of returning a `FileList`, it returns an Array of files explicitly due to some limitations.  I can't remember quite why off the top of my head, but maybe @nicolechung remembers and can chime in.

---

For some code golf, maybe:

```
typeof value === 'undefined' ||
        Array.isArray(value) ||
        (value as File[]).map((file) => file instanceof File).every((file) => Boolean(file))
```

---

Or maybe -- do we only care if they provide us an array or undefined and we don't need to get this granular with the types?

```
typeof value === 'undefined' || Array.isArray(value)
```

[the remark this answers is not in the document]
clintcs: I'd be curious to hear what @nicolechung says. The custom type (`File[]`) in lieu of its native representation is unfortunate.

Either way, if you go with what you have above, you could golf `typeof value === 'undefined'` down to `value === undefined`. And do you need to `map` and the cast? Or could you reduce it `(value).every((file) => file instanceof File))`?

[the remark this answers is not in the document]
ynotdraw: Doh, I'm dumb. You're right the single every will do!

packages/ember-toucan-core/src/components/form/controls/file-input.hbs
```diff
@@ -1,4 +1,5 @@
 <input
+  accept={{this.accept}}
```
simonihmig, line 2: :+1:

packages/ember-toucan-form/src/components/toucan-form.hbs
```diff
@@ -18,6 +18,9 @@
         (ensure-safe-component this.CheckboxGroupComponent) form=form
       )
       Field=form.Field
+      FileInput=(component
+        (ensure-safe-component this.FileInputFieldComponent) form=form
```
simonihmig, line 22: We should probably migrate our components to gts format (we only have this for tests I think), similar to https://github.com/CrowdStrike/ember-headless-form/pull/84, which (besides being nicer by itself) allows us to get rid of `ensure-safe-component`...

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
simonihmig: > but Array.prototype.at's return type includes undefined, which forces me to assert against undefined instead of possibly allowing a runtime error to propagate.

That's a good point! That really is a pretty bad mistake by TS to not do this for normal index access. Do you happen to know if there are plans to change this? It would probably be the right thing to do, although it would likely cause a ton of new type errors. But TS doesn't follow Semver anyway, so... 

> Though FileList is array-like and so doesn't support .at. It does support .item.

IIRC we chose to pass the files as `File[]`, so it _is_ an array if I am not mistaken...

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
simonihmig: Where is that `19` coming from? Is there a real file that we are using here, that has this size? Maybe add a comment to explain, as it was not obvious to me?

simonihmig (commented):
Code-wise this looks great!

However, when playing with the live link, I noticed an issue: when selecting a file and removing it afterwards, you cannot select the same file again! See this screen capture:

https://github.com/CrowdStrike/ember-toucan-core/assets/1325249/3638d163-1b14-435c-ba7d-e2987a84e792

Not sure where this comes from, might totally (most likely) be an upstream bug in toucan-core and not related to this PR, in which case you can consider this PR approved, and we can and should address this in a separate one.

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
vitch: > Do you happen to know if there are plans to change this?

It's already there ([since ts 4.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess)) behind a config flag (`noUncheckedIndexedAccess`). I'm pretty sure the safer version is the default in the `tsconfig.json` files recommended by `ember` (we don't have it on internally but IIRC it is on in the v2 addon blueprint which has caused confusion when people migrate to v2 addons).

ynotdraw:
@simonihmig a most excellent catch!  It is indeed coming from the `toucan-core` side.  I've got a fix in a downstream branch and will push here shortly.  The TL;DR is we are hitting this (https://stackoverflow.com/questions/39484895/how-to-allow-input-type-file-to-select-the-same-file-in-react-component) and the suggestion there seems to work (reset `target.value = ''`).  I just need to write up some tests and then will be good to go. Nice find!

simonihmig (approved):

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
ynotdraw: Good idea, I'll add a comment. Is indeed based on our test file we create. Alternatively we could NOT assert the size if we want, as the other assertions are probably fine enough?

test-app/tests/integration/components/toucan-form/toucan-form-test.gts
clintcs: > IIRC we chose to pass the files as File[], so it is an array if I am not mistaken...

Hm. If that's the reason, I wonder if we wouldn't be better off converting to an array as needed using `Array.from`.

packages/ember-toucan-core/src/components/form/controls/file-input.hbs
nicolechung: Thanks for adding this!

simonihmig: I wondered about this as well, but then found that this [already exists](https://github.com/CrowdStrike/ember-toucan-core/blob/64265d5bc77e2426c0152f9177bc4089f786f196/packages/ember-toucan-core/src/components/form/controls/file-input.ts#L38-L40), probably was just not used before, or lost during some refactoring/rebasing, idk?

ynotdraw: Yup! That's what I was thinking.  Maybe lost during refactoring or rebasing.  It is in the backing JS file, but it wasn't actually used anywhere (until now!)

packages/ember-toucan-form/src/components/toucan-form.hbs
ynotdraw: I agree!

packages/ember-toucan-core/src/components/form/controls/file-input.hbs
nicolechung: Ahh, thanks for linking me to it!

PR #178, opened 2023-06-13
chore: add @value support to Form::Controls::Checkbox

clintcs:
## 🚀 Description

This is a small but breaking change that removes `value` as an attribute in favor of `@value` the argument.  The reason for the change is consistency. Other controls that accept a value accept it by a `@value` argument.

---

## 🔬 How to Test

1. Navigate to the Checkbox Group Field at the preview URL.
2. Use your browser's inspector to inspect the `<input />` of one of the checkboxes. Note the value of its `value` property.
4. Confirm that value is equal to the `@value` specified in the demo template below it.

### DIFF
--- a/packages/ember-toucan-core/src/components/form/fields/checkbox.ts
+++ b/packages/ember-toucan-core/src/components/form/fields/checkbox.ts
@@ -79,7 +79,7 @@ export interface ToucanFormCheckboxFieldComponentSignature {
      *
      * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes
      */
-    value?: string;
+    value?: ToucanFormCheckboxControlComponentSignature['Args']['value'];
   };
   Blocks: {
     label: [];
--- a/.changeset/polite-bobcats-do.md
+++ b/.changeset/polite-bobcats-do.md
@@ -1,0 +1,5 @@
+---
+'@crowdstrike/ember-toucan-core': minor
+---
+
+add @value support to Form::Controls::Checkbox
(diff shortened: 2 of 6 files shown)


packages/ember-toucan-core/src/components/form/fields/checkbox.ts
```diff
@@ -79,7 +79,7 @@ export interface ToucanFormCheckboxFieldComponentSignature {
      *
      * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes
      */
-    value?: string;
+    value?: ToucanFormCheckboxControlComponentSignature['Args']['value'];
```
ynotdraw, line 82: 🎉

ynotdraw (commented):
Looks good to me!  You'll want to run:

```
pnpm changeset
```

at the root and pick a `minor` version change to add a release note log before merging

ynotdraw:
Not sure why, but changesets thinks `toucan-form` will have a major bump, which I think would bump us over to `1.0.0`. But looking at https://github.com/CrowdStrike/ember-toucan-core/pull/178/files#diff-323c7db4186b4b473c17647da304904a8437c66e93bea4f7edc814a2291efee5, there's only a line for `toucan-core`.  I'm not sure what Changesets is on about!

Screenshot 2023-06-13 at 3 20 19 PM

ynotdraw (approved):

simonihmig (approved):

simonihmig:
> Not sure why, but changesets thinks toucan-form will have a major bump

Weird indeed! 🤨

PR #183, opened 2023-06-14
File input field re-selection fix

ynotdraw:
## 🚀 Description

This PR fixes the issue @simonihmig mentioned at https://github.com/CrowdStrike/ember-toucan-core/pull/177#pullrequestreview-1478885953, where re-selecting the same file after a removal does nothing.

---

## 🔬 How to Test

- Go to https://0093993e.ember-toucan-core.pages.dev/docs/components/file-input-field
- Remove the existing file
- Click Select Files
- Add a file
- Remove the file
- Re-add the **same file you did previously**
- Verify the file was re-added

---

N/A

### DIFF
--- a/packages/ember-toucan-core/src/components/form/controls/file-input.ts
+++ b/packages/ember-toucan-core/src/components/form/controls/file-input.ts
@@ -68,4 +68,26 @@ export default class ToucanFormControlsFileInputComponent extends Component<Touc
 
     return this.args.onChange?.(files, event);
   }
+
+  /**
+   * As of 2023-06-15 on Chrome, when a file is selected in an input[type="file"]
+   * field multiple times change events are not triggered on the input.
+   *
+   * To get around that, we need to clear the input's value before the user
+   * selects the file again.
+   *
+   * This isn't a testable change because our test setup does not have high
+   * level user interactions such as "select file". To test this manually
+   * would require us to manually trigger the change event on the input field
+   * with the selected file(s), but that would not have caught the issue
+   * because we'd then need to *also* simulate the click event on the input
+   * as well.
+   *
+   * - https://stackoverflow.com/questions/39484895/how-to-allow-input-type-file-to-select-the-same-file-in-react-component
+   * - https://github.com/ngokevin/react-file-reader-input/issues/11#issuecomment-363484861
+   */
+  @action
+  resetValue(event: Event) {
+    (event.target as HTMLInputElement).value = '';
+  }
 }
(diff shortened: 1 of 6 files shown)


packages/ember-toucan-core/src/components/form/controls/file-input.ts
ynotdraw, line 1: For some reason I couldn't reproduce this issue in a test (by writing the test first) using:

```js
await triggerEvent('[data-file-input-field]', 'change', {
  files: [testFile],
});

// Delete the file

await triggerEvent('[data-file-input-field]', 'change', {
  files: [testFile],
});
```

I even tried moving this fix to the `onChange` action instead and nada.  I'm not so sure this is testable.  I may give this another shot in the AM though (with fresh eyes)

ynotdraw: This morning I wrote the following test on `main` to try and reproduce the bug:

```ts

  test('it handles the case where the same file is re-added', async function (assert) {
    const file = createFile(['Upload file sample'], {
      name: 'sample.txt',
      type: 'text/plain',
    });

    class Context {
      @tracked files: File[] = [];
    }

    let ctx = new Context();

    // Note: typescript can't tell the assert is checking the event
    // so it's complaining that the event is unused
    /* eslint-disable @typescript-eslint/no-unused-vars */
    const handleChange = (files: File[], event: FileEvent) => {
      ctx.files = files;
    };

    await render(<template>
      <FileInputField
        @deleteLabel="Delete File"
        @label="Label"
        @trigger="Select Files"
        @onChange={{handleChange}}
        @files={{ctx.files}}
        data-file-input-field
      />
    </template>);

    await triggerEvent('[data-file-input-field]', 'change', { files: [file] });

    // Verify files are there

    // Delete the file
    await triggerEvent('[data-delete-file]', 'click');

    // // Verify the `ul` is gone as all files are deleted
    assert.dom('[data-files]').doesNotExist();

    // Re-add the same file
    await triggerEvent('[data-file-input-field]', 'change', { files: [file] });

    // We expect this to fail if it acts just like the browser does IRL,
    // but it does not
  });

```

Unfortunately the last assertion I would expect to fail if it operates exactly as the browser does, but it does not. I don't think we can successfully test this functionality!  Totally open to suggestions or thoughts though.  Think adding an additional comment in the code about this not being testable would be sufficient?

clintcs (approved):

packages/ember-toucan-core/src/components/form/controls/file-input.ts
nicolechung: Based on our previous (offline) discussion it appears that this is untestable. I see that you copied those notes below!

simonihmig (approved):
:+1:

PR #190, opened 2023-06-21
Update qunit + test-helper dependencies

ynotdraw:
## 🚀 Description

This PR updates `ember-qunit` and `@ember/test-helpers` dependencies.  I noticed renovate was struggling with these ( [1](https://github.com/CrowdStrike/ember-toucan-core/pull/186), [2](https://github.com/CrowdStrike/ember-toucan-core/pull/167) ) since there were breaking changes, so figured I'd get things up to date.

---

## 🔬 How to Test

- Green build

---

N/A


packages/ember-toucan-core/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "peerDependencies": {
     "@crowdstrike/ember-toucan-styles": "^2.0.1",
-    "@ember/test-helpers": "^2.8.1",
+    "@ember/test-helpers": "^2.8.1 || ^3.0.0",
```
ynotdraw, line 36: I found it strange as to why this was listed as a `peerDependency`, but then stumbled upon https://github.com/CrowdStrike/ember-toucan-core/blob/main/packages/ember-toucan-core/src/test-support/components/button.ts (which I added at the time 🙈).  Due to us exposing some test helpers, we need this (unless we'd rather move it to a `dependency`, I suppose). 

We may also want to decide if we want to increase test-support in the future by providing more fractal-page-objects for our other components, now that we have quite a few.

docs-app/package.json
```diff
@@ -50,7 +50,6 @@
     "@tsconfig/ember": "^2.0.0",
     "@types/dompurify": "^3.0.0",
     "@types/ember": "^4.0.1",
-    "@types/ember-qunit": "^5.0.2",
```
ynotdraw, line 53: It appears we don't need these anymore with the latest version

packages/ember-toucan-core/package.json
simonihmig: Yeah, I think this seems correct then. You _could_ probably declare this peerDep as optional, on the other hand basically every app already has a dependency on `@ember/test-helpers`, so that doesn't matter much I guess...

simonihmig (approved):

docs-app/package.json
nicolechung: Sort of related: I think you can get rid of https://github.com/CrowdStrike/ember-toucan-core/blob/f1b73cd89c3f34319026ac7ed98d2304942c7a5d/docs-app/package.json#L71 (@types/ember__test-helpers) if you are using the most recent (ish) version of @ember/test-helpers

Rationale: https://www.npmjs.com/package/@types/ember__test-helpers

nicolechung (approved):

docs-app/package.json
ynotdraw: Excellent catch! Just pushed the update

PR #193, opened 2023-06-27
Add pnpm-sync-dependencies-meta-injected to improve DX

ynotdraw:
## 🚀 Description

This PR follows @vitch 's `headless-form` PR to make local development a lot better.  Instead of copying over what's there, I'm instead using @NullVoxPopuli 's [package](https://github.com/NullVoxPopuli/pnpm-sync-dependencies-meta-injected) that does this for us.

With these changes, instead of doing `pnpm` linking or running a script for local development, run `pnpm start` from the root of the monorepo. When making a change, the packages will rebuild and your change will be displayed.

- https://github.com/CrowdStrike/ember-headless-form/pull/140

---

## 🔬 How to Test

- Pull this branch down
- Run `pnpm i`
- Run `pnpm start`
- Verify the docs and test apps start
- Make a change under `ember-toucan-core` and `ember-toucan-form`
- Verify the change shows up in the docs/test app

---

## 📸 Images/Videos of Functionality

N/A - developer improvement only

PR #194, opened 2023-06-27
Fix: Apply text-disabled to labels and hint blocks when components are disabled

ynotdraw:
(Downstream of https://github.com/CrowdStrike/ember-toucan-core/pull/193 for DX improvement reasons)

## 🚀 Description

Adds disabled styling to labels and hint blocks when `@isDisabled`.

---

## 🔬 How to Test

View the following URLs and verify `text-disabled` is applied

- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/textarea-field#textareafield-with-label-and-isdisabled
- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/radio-group-field#radiogroupfield-with-label-and-isdisabled
- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/radio-field#radiofield-with-label-hint-and-value-isdisabled
- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/input-field#inputfield-with-label-and-isdisabled
- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/file-input-field#fileinputfield-with-label-and-isdisabled
- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/checkbox-group-field#checkboxgroupfield-with-label-and-isdisabled
- https://ceb26a4d.ember-toucan-core.pages.dev/docs/components/checkbox-field#checkboxfield-with-hint-and-isdisabled

---

| BEFORE  | AFTER |
| ------------- | ------------- |
| Screenshot 2023-06-27 at 7 37 00 PM  | Screenshot 2023-06-27 at 7 36 50 PM  |

### DIFF
--- a/packages/ember-toucan-core/src/-private/components/hint.ts
+++ b/packages/ember-toucan-core/src/-private/components/hint.ts
@@ -2,7 +2,12 @@ import templateOnlyComponent from '@ember/component/template-only';
 
 export interface ToucanFormHintComponentSignature {
   Element: HTMLDivElement;
-  Args: {};
+  Args: {
+    /**
+     * Sets disabled styling on the hint.
+     */
+    isDisabled?: boolean;
+  };
   Blocks: {
     default: [];
   };
--- a/packages/ember-toucan-core/src/components/form/fields/checkbox-group.hbs
+++ b/packages/ember-toucan-core/src/components/form/fields/checkbox-group.hbs
@@ -16,7 +16,8 @@
       )
     }}
       <legend
-        class="type-md-tight text-body-and-labels flex items-center gap-1.5"
+        class="type-md-tight flex items-center gap-1.5
+          {{if @isDisabled 'text-disabled' 'text-body-and-labels'}}"
         data-label
       >
         {{#if (has-block "label")}}
@@ -36,7 +37,7 @@
         (hash blockExists=(has-block "hint") argName="hint" arg=@hint)
       )
     }}
-      <field.Hint id={{field.hintId}} data-hint>
+      <field.Hint id={{field.hintId}} data-hint @isDisabled={{@isDisabled}}>
         {{#if (has-block "hint")}}
           {{yield to="hint"}}
         {{else}}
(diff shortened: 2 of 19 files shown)


packages/ember-toucan-core/src/components/form/fields/checkbox-group.hbs
```diff
@@ -16,7 +16,8 @@
       )
     }}
       <legend
-        class="type-md-tight text-body-and-labels flex items-center gap-1.5"
+        class="type-md-tight flex items-center gap-1.5
+          {{if @isDisabled 'text-disabled' 'text-body-and-labels'}}"
```
ynotdraw, line 20: It may be tempting to make a `GroupLegend` component or something, that's essentially `Label` but renders a `<legend` and can be used here and with the radio-group; however, I punted on this for now.  Happy to do it in a follow up if folks think it'd be worth it though.

test-app/tests/integration/components/checkbox-field-test.gts
```diff
@@ -15,6 +15,7 @@ module('Integration | Component | Fields | CheckboxField', function (hooks) {
     </template>);
 
     assert.dom('[data-label]').hasText('Label');
+    assert.dom('[data-label]').hasClass('text-titles-and-attributes');
```
ynotdraw, line 18: I've got mixed feelings on asserting class names in tests, as I'd prefer visual regression testing instead; however, I want to make sure we don't regress **until** we get something like Percy added.

test-app/tests/integration/components/file-input-field-test.gts
```diff
@@ -57,6 +57,8 @@ module('Integration | Component | Fields | FileInput', function (hooks) {
     </template>);
 
     assert.dom('[data-label]').hasText('Label');
+    assert.dom('label').hasClass('text-body-and-labels');
```
ynotdraw, line 60: FileInput is a bit different than the others, where because of the markup, `data-label` is actually a nested element within `<label`.  Due to that, I'm targeting the `label` tag directly here instead of using `[data-label]`

packages/ember-toucan-core/src/-private/components/hint.ts
```diff
@@ -2,7 +2,12 @@ import templateOnlyComponent from '@ember/component/template-only';
 
 export interface ToucanFormHintComponentSignature {
   Element: HTMLDivElement;
-  Args: {};
+  Args: {
+    /**
+     * Should the hint text be styled in a disabled state?
```
clintcs, line 7: Nit: This reads like a todo to me.

clintcs (approved):

PR #200, opened 2023-07-05
[WIP] || Add Combobox component

ynotdraw:
# NOTE: This is incredibly rough. Please do not review yet.

## 🚀 Description

---

## 🔬 How to Test

---


packages/ember-toucan-core/src/-private/components/form/controls/combobox/option.hbs
```diff
@@ -0,0 +1,36 @@
...
+  {{! template-lint-disable no-nested-interactive }}
+  {{! template-lint-disable require-input-label }}
+  <input
+    class="hidden"
+    checked={{@isSelected}}
+    disabled={{@isDisabled}}
+    readonly={{@isReadOnly}}
+    type="checkbox"
+    value={{@value}}
+  />
```
ynotdraw, line 35: I'm a bit unclear on how we feel about this as a group.  Clinton added this super helpful comment about FormData.  I don't have any strong opinions leaving this in here or ripping it out.  What do y'all think?

packages/ember-toucan-form/src/-private/combobox-field.ts
```diff
@@ -0,0 +1,72 @@
...
+     * @internal
+     */
+    form: HeadlessFormBlock<DATA>;
+  };
+  // TODO: How do we get this to play nicely with our
+  // generic in toucan-core?
+  // `BaseComboboxFieldSignature<Option>['Blocks'];`
+  // gives a glint error!
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Blocks: BaseComboboxFieldSignature<any>['Blocks'];
```
ynotdraw, line 36: I couldn't get this to work as expected unfortunately so I made the type `any`.  We want the blocks to take a generic like the Combobox does, but since this component is already generic due to the headless-form data, I'm not sure how to make this match the typing of `@options` like we do above! Any thoughts?  Maybe @simonihmig ?

To reproduce you can pull this repo down and the test at https://github.com/CrowdStrike/ember-toucan-core/pull/200/files#diff-a7fdb0c7554c73ada47e314a1462054c5557c84b0bdf6649337495dc0cc96c5eR31 should throw an error if you change this from `any` to something else.

test-app/package.json
ynotdraw, line 1: `pnpm` apparently decided to reorder `ember-source` 🤷

packages/ember-toucan-core/src/-private/icons/lock.ts
ynotdraw, line 1: Clinton suggested making `-private/icons` instead of icons living at `/-private/components`.  I liked the idea, so this is why it changed!

ynotdraw:
During a discussion today, it was suggested we maybe rename this component to `Autocomplete` to match [mui](https://mui.com/material-ui/react-autocomplete/).  I'll do this on Monday!

packages/ember-toucan-core/package.json
```diff
@@ -43,13 +43,17 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.20.7",
-    "@embroider/addon-shim": "^1.0.0"
+    "@embroider/addon-shim": "^1.0.0",
+    "@embroider/compat": "^2.0.0",
```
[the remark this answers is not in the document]
simonihmig: I have no idea what the error means, but depending on `@embroider/compat` is definitely not the right thing to do! 

We will need to find the root cause, but let's remove that dependency anyway...

test-app/tests/integration/components/combobox-field-test.gts
```diff
@@ -0,0 +1,247 @@
...
+    </template>);
+
+    let error = find('[data-error]');
+
+    assert.dom(error).hasText('Error text');
+    assert.dom(error).hasAttribute('id');
+
+    let errorId = error?.getAttribute('id') || '';
+
+    assert.ok(errorId, 'Expected errorId to be truthy');
```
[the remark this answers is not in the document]
ynotdraw: Yeah, if you do `assert.ok('')` it will fail, as `''` is falsy!  `assert.ok` expects the value to be truthy

test-app/tests/integration/components/combobox-field-test.gts
```diff
@@ -0,0 +1,247 @@
...
+      );
+
+    assert.dom('[data-combobox]').hasTagName('input');
+    assert.dom('[data-combobox]').hasAttribute('id');
+    assert.dom('[data-combobox]').hasClass('text-titles-and-attributes');
+
+    assert
+      .dom('[data-error]')
+      .doesNotExist(
+        'Expected hint block not to be displayed as an error was not provided'
```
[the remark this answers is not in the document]
ynotdraw: Just pushed the suggestion!

test-app/tests/integration/components/combobox-field-test.gts
```diff
@@ -0,0 +1,247 @@
...
+
+    assert.dom('[data-lock-icon]').exists();
+  });
+
+  test('it sets readonly on the textarea using `@isReadOnly` and renders a lock icon', async function (assert) {
+    await render(<template>
+      <ComboboxField @label="Label" @isReadOnly={{true}} data-combobox />
+    </template>);
+
+    assert.dom('[data-combobox]').hasAttribute('readonly');
```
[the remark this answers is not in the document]
ynotdraw: I personally think testing for the attribute is enough.  Since we are using the platform™, we can trust that the `readonly` attribute operates as defined at https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

test-app/tests/integration/components/combobox-test.gts
```diff
@@ -0,0 +1,854 @@
...
+    assert.dom('[data-combobox]').doesNotHaveClass('text-disabled');
+    assert.dom('[data-combobox]').doesNotHaveClass('shadow-error-outline');
+    assert
+      .dom('[data-combobox]')
+      .doesNotHaveClass('focus:shadow-error-focus-outline');
+
+    assert.dom('[data-combobox]').hasAttribute('aria-autocomplete', 'list');
+    assert.dom('[data-combobox]').hasAttribute('aria-haspopup', 'listbox');
+    assert.dom('[data-combobox]').hasAttribute('autocapitalize', 'none');
+    assert.dom('[data-combobox]').hasAttribute('autocomplete', 'off');
```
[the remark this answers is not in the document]
ynotdraw: Good question!

It is a little confusing, especially since we are eventually wanting to rename this to "autocomplete" instead.  We want to turn this attribute `off`, as this is the browser's default interactions as defined at https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete.

> The browser is not permitted to automatically enter or select a value for this field. It is possible that the document or application provides its own autocomplete feature, or that security concerns require that the field's value not be automatically entered.

That last part is the important piece - we are defining our own autocomplete feature via this component. By us setting `autocomplete="off"` here, we are telling the browser **not** to suggest autocomplete options.  Rather, _we_ provide the suggestions manually via the popover contents.  

Maybe as a TL;DR: we want to disable the browser's built-in autocomplete and use our own implementation instead.

docs/components/combobox/index.md
```diff
@@ -0,0 +1,254 @@
...
+Provides a Toucan-styled combobox with filtering.
+If you are building forms, you may be interested in the ComboboxField component instead.
+
+## Content Class
+
+A CSS class to add to this component's content container. Commonly used to specify a `z-index`.
+
+```hbs
+<Form::Controls::Combobox @contentClass='z-50' />
+```
```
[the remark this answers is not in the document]
nicolechung: It's too bad we can't (selectively) expose the `Combobox.Content/Popover` and let folks style it via `...attributes`.

Seems like the way `--z-index-popover` is set is via a `:root` in users css files...so I'm leaning towards a combination of 1 and 3, if that sets a sensible default that works for most people and then gives them ways to override it if needed.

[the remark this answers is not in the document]
ynotdraw: Cool! Went ahead and added options 1 & 3.  Thanks, Nicole!

docs/components/combobox/index.md
```diff
@@ -0,0 +1,254 @@
...
+  options = ['Blue', 'Red', 'Yellow'];
+
+  @action
+  handleChange(option) {
+    this.selected = option;
+  }
+}
+```
+
+## Option Key
```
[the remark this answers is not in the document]
nicolechung: Are `options` going to have other ... options other than `@optionKey` in the future? Just wondering if it's worthwhile to think about putting that into a grab bag of @options (containing the options array, the optionKey, and future stuff)

packages/ember-toucan-core/src/-private/components/form/controls/combobox/option.hbs
nicolechung: I like the super helpful comment. I'd prefer `<input>` or `<input />` over `INPUT` because I was thinking that `INPUT` refers to a special `INPUT` component we've created (got distracted with how it's written). Also maybe something about how a `hidden` input is required to satisfy the FormData API (users don't need to see the input, the FormData api does)

Maybe that's just me tho...I had to read this a few times to connect all the parts.

docs/components/combobox-field/index.md
```diff
@@ -0,0 +1,311 @@
+# Combobox Field
+
+Provides a Toucan-styled combobox with filtering that builds on top of the Field component.
```
nicolechung, line 3: Suggestions: Link to the field component, and maybe mention the benefits (error, labels, hint)

test-app/tests/integration/components/combobox-test.gts
```diff
@@ -0,0 +1,854 @@
...
+      </ComboboxControl>
+    </template>);
+
+    // Open the popover by clicking it
+    await click('[data-combobox]');
+
+    // Now close it
+    await triggerKeyEvent('[data-combobox]', 'keydown', 'Escape');
+    // Now reopen it
+    await triggerKeyEvent('[data-combobox]', 'keydown', 'ArrowDown');
```
nicolechung, line 548: Super tiny nit / question: you can also reopen it with `Escape`, is this intentional? I actually don't mind this experience tho (less having to move around the keyboard)

test-app/tests/integration/components/combobox-test.gts
```diff
@@ -0,0 +1,854 @@
...
+
+    assert.dom('[role="listbox"]').exists();
+  });
+
+  test('it makes the first option "active" when the metakey and UP arrow is pressed', async function (assert) {
```
nicolechung, line 553: Might be worth pointing out that metakey is CMD in Mac (not sure what it is in Windows, CTRL?)

nicolechung: also `active` is the "currently hovered" one in this case? you still have to press ENTER to select? (just to clarify)

test-app/tests/integration/components/combobox-test.gts
```diff
@@ -0,0 +1,854 @@
...
+      .hasAttribute('data-active', 'false');
+  });
+
+  /**
+   * This test covers the case where we have a selected option,
+   * the user clears the input completely, and then the component's
+   * input tag is blurred.  In that case, we want to clear the input
+   * and reset their selection to null.
+   */
+  test('it calls `@onChange` with null after a user clears their original input selection', async function (assert) {
```
nicolechung, line 710: Could just be me missing this...but is there a test that verifies that when a user focuses on the combobox input when text is already there, that it selects ALL test?

test-app/tests/integration/components/combobox-test.gts
```diff
@@ -0,0 +1,854 @@
...
+    assert.verifySteps([]);
+
+    // We want to verify the original options are re-displayed
+    // rather than the input being filtered to garbage
+    await click('[data-combobox]');
+
+    assert.dom('[role="option"]').exists({ count: 2 });
+  });
+
+  test('it reverts to the selected option when a user enters garbage after previously having a valid selection (with `@selected` as an object)', async function (assert) {
```
nicolechung, line 794: 🎉  Love these exhaustive tests! Really gives me a lot of confidence in how to understand how this feature works.

nicolechung (approved):
I really like how it selects all text when you focus back on the field.

I've asked @noahgelmancs  to take a closer look at the a11y tests to give a 👍 

This looks like a HUGE amount of work 👏  very excited to see this happening!

docs/components/combobox-field/index.md
ynotdraw: Mind if I do this in a follow up PR? This is copy/pasta'd from the other docs, so I'd like to do them all in one go so it's consistent if possible!

test-app/tests/integration/components/combobox-test.gts
ynotdraw: I'm not sure if we can test this! Do you know how to test what is currently selected?  I was under the impression there's not a way to do this with qunit, but I could be wrong!

test-app/tests/integration/components/combobox-test.gts
ynotdraw: > also active is the "currently hovered" one in this case? you still have to press ENTER to select? (just to clarify)

Correct!

test-app/tests/integration/components/combobox-test.gts
ynotdraw: - [ ] Let me try `document.getSelection().toString()` and see if that'll work!

test-app/tests/integration/components/combobox-test.gts
ynotdraw: Yeah, it is intentional, but not sure if it's "correct".  I thought any keyboard input opening the popover was suitable UX, but that's maybe just a personal preference.  I'm glad you don't mind it either!

test-app/tests/integration/components/combobox-test.gts
ynotdraw: Good suggestion! That worked! Test added.

test-app/tests/integration/components/combobox-test.gts
```diff
@@ -851,4 +851,31 @@ module('Integration | Component | Combobox', function (hooks) {
 
     assert.dom('[role="option"]').exists({ count: 2 });
   });
+
+  test('it reselects the entered input value when there is a selected item and the input regains focus', async function (assert) {
```
nicolechung, line 855: 🎉

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+    assert(
+      '`this.args.options` cannot be  `undefined`',
+      this.args.options !== undefined
+    );
+
+    let { optionKey, onChange } = this.args;
+
+    this.closePopover();
+
+    // This shouldn't be possible, but to satisfy TS
```
simonihmig, line 252: Nit: If it's not possible to be in that state, then we can use `assert()`, which makes TS happy and does not add runtime code (in prod).

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+    owner: unknown,
+    args: ToucanFormComboboxControlComponentSignature<OPTION>['Args']
+  ) {
+    super(owner, args);
+
+    // We need to set our input tag's value attribute
+    // if we have a selected option provided on render
+    let { selected, optionKey } = this.args;
+
+    this.inputValue =
```
simonihmig, line 125: I was expecting `inputValue` to be derived state. Do we really need to handle that state imperatively? Basically the logic below, but as a getter...

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+
+    /**
+     * The function called when a user types into the combobox textbox, typically used to write custom filtering logic.
+     */
+    onFilter?: (input: string) => OPTION[];
+
+    /**
+     * `@options` forms the content of this component.
+     *
+     * To support a variety of data shapes, `@options` is typed as `unknown[]` and treated as though it were opaque.
```
simonihmig, line 65: AFAICT, it is typed as `Array<string | Record<string, unknown> | undefined>` though, not unknown! Which is good I think, but the docs are misleading here.

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+     */
+    options?: OPTION[];
+
+    /**
+     * When `@options` is an array of objects, `@selected` is also an object.
+     * The `@optionKey` is used to determine which key of the object should
+     * be used for both filtering and displayed the selected value in the
+     * textbox.
+     */
+    optionKey?: string;
```
simonihmig, line 76: We maybe able to type this in a more strict (correct) way:

```suggestion
    optionKey?: OPTION extends Record<string, unknown> ? keyof OPTION : undefined;
```

I haven't tried this, so maybe it doesn't work or causes issues. But given...
* `optionsOfObjects = [{label: 'foo'}, {label: 'bar'}]`
* `optionsOfStrings = ['foo' 'bar']`

ideally TS would give us this feedback...
* `@options={{optionsOfObjects)} @optionKey="label"` ✅
* `@options={{optionsOfStrings}}` ✅
* `@options={{optionsOfObjects)} @optionKey="title"` ❌
* `@options={{optionsOfStrings)} @optionKey="label"` ❌
* `@options={{optionsOfObjects)}` ❌

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+      this.inputValue = selectedOption;
+    }
+
+    if (selectedOption && typeof selectedOption === 'object' && optionKey) {
+      let option = (selectedOption as Record<string, string>)[optionKey];
+
+      this.inputValue = option;
+    }
+
+    this.args.onChange?.(this.options[this.activeIndex]);
```
simonihmig, line 278: Normally the user would set whatever gets passed to `@selected` to this new option, but what happens when that does not happen? Maybe there are reasons not to update, like it's an invalid value or something.

I think what happens now is that `inputValue` represents the new option, while the state of `@selected` is still the old one. Which would be inconsistent state, right? 

I think that wouldn't happen if `inputValue` was derived state, as suggested above...

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+
+    return (
+      typeof selected === 'object' && optionKey ? selected[optionKey] : selected
+    ) as string;
+  }
+
+  /**
+   * Attempts to scroll the active or newly highlighted item into view for the user.
+   */
+  #scrollActiveOptionIntoView(alignToTop?: boolean) {
```
simonihmig, line 211: Why a private member here? I mean it's not wrong, but the (Glimmer) component model is basically that _all_ props and methods are private effectively, as you can only pass args from the outside. So seems inconsistent, and not really needed here...

packages/ember-toucan-form/src/-private/combobox-field.ts
simonihmig: I think we would have to introduce `OPTION` here also, and pass it from `ToucanFormComboboxFieldComponent` to `ToucanFormComboboxFieldComponentSignature` down to `BaseComboboxFieldSignature`. have you tried that?

test-app/config/ember-try.js
```diff
@@ -41,30 +43,32 @@ module.exports = async function () {
...
-          devDependencies: {
-            'ember-source': releaseVersion,
-            // @todo remove this once we have a stable release that includes https://github.com/embroider-build/embroider/pull/1383, which should be https://github.com/embroider-build/embroider/pull/1408
-            '@embroider/core': '2.1.2-unstable.c58f146',
-            '@embroider/compat': '2.1.2-unstable.c58f146',
-            '@embroider/webpack': '2.1.2-unstable.c58f146',
-          },
-        },
-      }),
+      // @todo Temporarily disabling embroider optimized tests due to https://github.com/CrowdStrike/ember-toucan-core/issues/210
```
simonihmig, line 46: Nit: would have been enough to just not comment out the embroider scenarios in `ci.yml` as you did, and leave this as-is...

test-app/tests/integration/components/combobox-field-test.gts
```diff
@@ -0,0 +1,247 @@
...
+
+    assert.dom('[data-lock-icon]').exists();
+  });
+
+  test('it spreads attributes to the underlying textarea', async function (assert) {
```
simonihmig, line 144: testarea? 🙃

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+    /**
+     * Called when the user makes a selection.
+     * It is called with the selected option (derived from `@options`) as its only argument.
+     */
+    onChange?: (option: unknown) => void;
+
+    /**
+     * The function called when a user types into the combobox textbox, typically used to write custom filtering logic.
+     */
+    onFilter?: (input: string) => OPTION[];
```
simonihmig, line 60: I saw we are calling this with ´await`, so should be this?

```suggestion
    onFilter?: (input: string) => OPTION[] | Promise<OPTION[]>;
```

packages/ember-toucan-core/src/components/form/controls/combobox.ts
```diff
@@ -0,0 +1,537 @@
...
+    if (!onFilter && optionKey) {
+      filteredOptions = optionsArgument?.filter((option) =>
+        ((option as Record<string, unknown>)[optionKey] as string)
+          ?.toLowerCase()
+          ?.startsWith(value.toLowerCase())
+      );
+    }
+
+    if (onFilter) {
+      filteredOptions = await onFilter(value);
```
simonihmig, line 438: Should we also pass the raw options here, so the function can be simple and not need to have a of closure to have access to the options? Like

```suggestion
      filteredOptions = await onFilter(optionsArgument, value);
```
Another option (pun intended) would be to make this function a higher order function that returns the filter function, so you could do:

```suggestion
      const filterFn = onFilter(value);
      filteredOptions = optionsArgument.filter(filterFn);
```

We lose the ability for async though. But it would have the benefit that you can only return options that are actually members of `@options`. Which would be weird if that wasn't the case, and likely cause problems (e.g. when doing `===` checks). In the suggestion above such a mistake would be less likely, and more likely in the original implementation here. 

I don't have strong opinions on this though...

docs/components/combobox/index.md
```diff
@@ -0,0 +1,254 @@
...
+  ];
+
+  @action
+  handleChange(option) {
+    this.selected = option;
+  }
+}
+```
+
+## onFilter
```
simonihmig, line 162: This can be marked as optional, right? Maybe also point out what the default filtering logic does, so it's clear you don't need this if the defaults work as well.

simonihmig (approved):
Approved, with some suggestions and remarks. I'm totally fine with either dismissing them if you disagree, or tackling them in future iterations, as I don't want to block this big PR any further...

simonihmig:
My earlier comment here was a bit unemotional or sober, and concentrated only on possible improvements without stating the most important thing: this is some absolutely beautiful work here! Congrats @ynotdraw and @clintcs! 🎉 😅

docs/components/combobox/index.md
```diff
@@ -0,0 +1,254 @@
+# Combobox
+
+Provides a Toucan-styled combobox with filtering.
+If you are building forms, you may be interested in the ComboboxField component instead.
+
+## Popover z-index
+
+A CSS class to add to this component's content container. Commonly used to specify a `z-index`.
```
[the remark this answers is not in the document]
ynotdraw: More info over on https://github.com/CrowdStrike/ember-toucan-core/pull/200#discussion_r1256222867, but _hopefully_ this is temporary and we can remove it soon™.  We just need to figure out where/if some of the CSS variables we have should be defined!

docs/components/combobox/index.md
```diff
@@ -0,0 +1,254 @@
...
+```hbs
+<Form::Controls::Combobox
+  @onChange={{this.handleChange}}
+  @options={{this.options}}
+  @optionKey='label'
+  @selected={{this.selected}}
+  as |combobox|
+>
+  <combobox.Option>
+    {{combobox.option}}
```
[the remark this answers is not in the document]
ynotdraw: Excellent catch! Just committed the suggestion.

packages/ember-toucan-core/src/-private/components/form/controls/combobox/option.hbs
```diff
@@ -0,0 +1,36 @@
...
+    {{this.styles}}
+    {{this.className}}
+    "
+  data-active={{if @isActive "true" "false"}}
+  id="{{@popoverId}}-{{@index}}"
+  role="option"
+  {{on "click" this.onClick}}
+  {{! template-lint-disable no-pointer-down-event-binding }}
+  {{on "mousedown" this.onMousedown}}
+  {{on "mouseover" @onMouseover}}
```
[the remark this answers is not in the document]
ynotdraw: Good callout - let me move this! 🙇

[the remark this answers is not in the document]
ynotdraw: Good question!  Yeah, I think we probably could.

This gets used in `combobox.hbs` where we yield back `Option`.  We do the following:

```hbs
onMouseover=(fn this.onOptionMouseover index)
```

Is there a way to make this a modifier rather than a component argument when we are yielding back?

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: Perfect! Pushed as part of commit https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/4f2d139737aa6ddb6025150574247a191cc1fe03

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: Doh, nice catch.  This was left over from when we previously had `unknown[]` rather than `Array<string | Record<string, unknown> | undefined>`.  Fixed with commit: https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/45594a7f284e0d3e2105254365d3e2d0c3695fe9

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: Let me give this a shot!

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: It works! This is great! Thanks a bunch, Simon!

Screenshot 2023-07-14 at 8 36 29 AM

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: Removed as part of https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/754b29b79b808728f95bcc2885851e362839c53b!

test-app/config/ember-try.js
ynotdraw: Fixed with https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/0c8f95813e2019c0f55897b9ada00b61b6673916

test-app/tests/integration/components/combobox-field-test.gts
ynotdraw: Thank you! Turns out a lot of copy pasta (https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/ac3991c03106e6b51f6cbea1f01b89623b4d0aab)

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: Ah yeah, I backed out of it being a promise for now due as the current implementation is not promise based.  Let me revert the `await`!  We can add that support back if we find out folks need it! https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/d036a432a1a36b293204e16f34cc57f437185cb5

docs/components/combobox/index.md
ynotdraw: Just wrote some more as part of https://github.com/CrowdStrike/ember-toucan-core/pull/200/commits/2b74de095b48c18335903e0aabaa25d5717cc734, thank you!

docs-app/package.json
```diff
@@ -135,6 +135,7 @@
     "ember-browser-services": "^4.0.4",
     "ember-modifier": "^4.1.0",
     "ember-resources": "^6.0.0",
+    "ember-velcro": "^2.1.0",
```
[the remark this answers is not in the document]
ynotdraw: Yeah, unfortunately without it we get the following error on a build: https://github.com/CrowdStrike/ember-toucan-core/actions/runs/5554403233/jobs/10144183085?pr=200

```bash
docs-app:build:   - originalErrorMessage: Module not found: Error: Can't resolve '../../../node_modules/.pnpm/ember-velcro@2.1.0_@babel+core@7.20.12_ember-source@5.1.0/node_modules/ember-velcro/components/velcro/index' in '$TMPDIR/embroider/e83321/docs-app/components/velcro/index.js'
```

😭

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: I'm going to keep playing with this in a downstream PR if that's okay so that the changes can be focused rather than keep adding onto the pile here

packages/ember-toucan-core/src/components/form/controls/combobox.ts
ynotdraw: Yeah for sure, I could see this being helpful.  I'm going to go ahead and merge as-is if that's okay due to our protection with changesets and am going to chat up @clintcs on Monday about this!

packages/ember-toucan-core/src/components/form/controls/combobox.ts
clintcs: Just came across this! 

It was my doing. I tend to make private things private regardless of how the framework treats them. But, even though Glimmer handles privacy for us, I still think private methods are useful in that they imply a method is not used in the template.

Or is the fact that the a private method can't be used in the template just a shortcoming of the template syntax and not by design?

packages/ember-toucan-core/src/components/form/controls/combobox.ts
clintcs: This resonated with me at first, but now I find myself questioning it—for better or worse 😆. 

One thing is I'm struggling to imagine a case where a consumer would show an invalid value. 

I also wonder how it would affect the user if the consumer chose not to set `@selected` to the selected value. If the user clicks, say, "Blue" and `@onChange` is called but the consumer decides not to update `@selected`, how should the UI respond? As is, the popover will be closed. But is that the right behavior? How should we let the user know that what he selected is invalid? And if we derive `this.inputValue`, what should its value be?

Part of me thinks that, to prevent all of this, the component shouldn't offer a `@selected` argument and should handle it internally. We could instead offer an `@initialSelected` argument, letting the consumer specify an initially selected state.

clintcs: @simonihmig: So you're up to date, @ynotdraw and I met up to discuss this. 

We realized that my suggestion of `@initialSelected` won't fly if the consumer decides to change `@options` after initial render. There's nothing to stop the consumer from changing `@options` to an entirely different set of options after initial render. And, if they do that, they may also want to update `@initialSelected`. But they won't be able to given we'd only be assigned it to `this.selected` on render. So we took `@initialSelected` off the table.

We also mulled over my user experience questions above and your initial concern about `this.inputValue`. In the end, we decided that we're okay with possibility of `this.inputValue` diverging from `@selected` in the case where the consumer decides not to update `@selected` after `@onChange`. We're okay with this in part because we couldn't imagine a reason why the consumer wouldn't want to update `@selected` and in part of because of the user experience complications I brought up above. 

@ynotdraw brought up another really good point: most consumers (we think) will consume Autocomplete (née Combobox) via the Form component. That means that most consumers won't be setting `@selected` themselves. Form will set it for them.

Let us know what you think? If you can think of a case where the consumer wouldn't want to set `@selected` after `@onChange`, we'd love to here it as well.

PR #205, opened 2023-07-07
pnpm 8, address some compatibility issues

NullVoxPopuli:
## 🚀 Description

---

## 🔬 How to Test

---


NullVoxPopuli:
>  CI / Publish Docs to Cloudflare Pages (pull_request) Failing after 1m 

Failures here will always fail from forks, unless that/those jobs are moved to a trusted workflow, like the strategy here:
- https://github.com/universal-ember/ember-primitives/blob/main/.github/workflows/deploy-preview.yml

PR #212, opened 2023-07-14
fix: button flex grow

joelamb:
## 🚀 Description
If the button is the child of a flex parent then,
when disabled, the yielded content wrapping span
needs a flex-grow CSS property to ensure that the
label can expand, and push the button width out
to fill the full width of the parent

---

## 🔬 How to Test

---

### DIFF
--- a/packages/ember-toucan-core/src/components/button.hbs
+++ b/packages/ember-toucan-core/src/components/button.hbs
@@ -9,7 +9,7 @@
     {{yield to="loading"}}
     <span class="sr-only" data-loading>{{yield}}</span>
   {{else if @isDisabled}}
-    <span class="flex items-center gap-x-2">
+    <span class="flex flex-grow items-center gap-x-2">
       {{yield}}
       {{yield to="disabled"}}
     </span>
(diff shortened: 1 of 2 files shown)


ynotdraw (approved):

ynotdraw (changes requested):
You'll want to add a changeset so this gets picked up in the release by running `pnpm changeset` at the root and choose a patch update (don't pick major or minor, just let it default at the end)!  Other than that looks great!

nicolechung:
Checked against https://ember-toucan-core.pages.dev/docs/components/button, don't see any regressions. Please add a changeset tho (I forget this often myself)

ynotdraw (approved):

packages/ember-toucan-core/src/components/button.hbs
```diff
@@ -9,7 +9,7 @@
     {{yield to="loading"}}
     <span class="sr-only" data-loading>{{yield}}</span>
   {{else if @isDisabled}}
-    <span class="flex items-center gap-x-2">
+    <span class="flex flex-grow items-center gap-x-2">
```
nicolechung, line 12: Based on your description, does the `flex-grow` only need to be here when the button is disabled?

joelamb: that's where I've encountered the error

PR #226, opened 2023-07-21
chore: Autocomplete tweaks

clintcs:
### Only accept an array of strings for `@options`

The consumer is currently allowed to pass either an array of strings _or_ an array of objects. @ynotdraw and I talked about it. We think removing support for arrays of objects makes for an overall better component. 

This change does move some work to the consumer. If the consumer's data structure is anything other than an array of strings, the consumer will now have to do a lookup after `@onChange(selected: string)` is called to find the selected object in the `@options` array. Consumers will also have to map their data to our `string[]` structure, though this is easily done.

Despite this additional work for the consumer, we think that the **disadvantages** of supporting an array of objects make this change worthwhile:

#### The implementation is more complex

Various [conditionals](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-837203e404c8fe1aa1e4644e395e8232a778a58071bc041170df58b22e60eeebL126) and [casts](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-837203e404c8fe1aa1e4644e395e8232a778a58071bc041170df58b22e60eeebL256) [were added throughout](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-f56c38f7ff91374f94b1488051f8a53a8b5d006e36303ced4c084d3e697ee333L70) to support the two data structures.

#### The API is more complex

 A generic (`OPTION`) was [added](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-837203e404c8fe1aa1e4644e395e8232a778a58071bc041170df58b22e60eeebL105) to (I believe) support `Form`. @ynotdraw can give you details if you're curious.

The [`@optionKey`](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-837203e404c8fe1aa1e4644e395e8232a778a58071bc041170df58b22e60eeebL75) argument was added and required to be passed by the consumer when `@options` is an object.

#### The slope is slippery

What about consumers who want to supply an array of objects where the desired data doesn't reside on the top level of each object in the array? Or maybe the consumer has an array of arrays or a mix of arrays and objects? 

The permutations are endless, and we're bound to encounter all of them given enough consumers. Thus supporting an array of objects isn't just that—it's supporting arbitrary data structures, which is not the business we want to be in.

### Various tweaks and nitpicks

🤷‍♂️

## 🔬 How to Test (Copied from the [original PR](https://github.com/CrowdStrike/ember-toucan-core/pull/200))

Navigate to [https://ebf80b3e.ember-toucan-core.pages.dev/docs/components/autocomplete](https://ebf80b3e.ember-toucan-core.pages.dev/docs/components/autocomplete)

- Mouse
  1. Click the Colors select control
  1. Verify the list of options opens and is displayed
  1. Type b
  1. Verify only blue is shown
  1. Click "Blue"
  1. Verify the popover closes and "Blue" is selected

- Keyboard
  1. Click the Colors select control
  1. Verify the list of options opens and is displayed
  1. Type b
  1. Verify only blue is shown
  1. Press the ENTER key
  1. Verify the popover closes and "Blue" is selected

### DIFF
--- a/.changeset/smart-mice-crash.md
+++ b/.changeset/smart-mice-crash.md
@@ -1,0 +1,6 @@
+---
+'@crowdstrike/ember-toucan-core': minor
+'@crowdstrike/ember-toucan-form': minor
+---
+
+Removed from Autocomplete support for `@options` as an array of objects.
--- a/packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
+++ b/packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
@@ -16,11 +16,7 @@ import Chevron from '../../../-private/icons/chevron';
 import type { Middleware as VelcroMiddleware } from '@floating-ui/dom';
 import type { WithBoundArgs } from '@glint/template';
 
-export type Option = string | Record<string, unknown> | undefined;
-
-export interface ToucanFormAutocompleteControlComponentSignature<
-  OPTION extends Option
-> {
+export interface ToucanFormAutocompleteControlComponentSignature {
   Args: {
     /**
      * A CSS class to add to this component's content container.
@@ -52,39 +48,31 @@ export interface ToucanFormAutocompleteControlComponentSignature<
      * Called when the user makes a selection.
      * It is called with the selected option (derived from `@options`) as its only argument.
      */
-    onChange?: (option: unknown) => void;
+    onChange?: (selected: string | null) => void;
 
     /**
      * The function called when a user types into the textbox, typically used to write custom filtering logic.
+     *
+     * @return An array of options to be shown in the popover.
      */
-    onFilter?: (input: string) => OPTION[];
+    onFilter?: (value: string) => string[];
 
     /**
      * `@options` forms the content of this component.
      *
      * `@options` is simply iterated over then passed back to you as a block parameter (`select.option`).
      */
-    options?: OPTION[];
+    options?: string[];
 
     /**
-     * When `@options` is an array of objects, `@selected` is also an object.
-     * The `@optionKey` is used to determine which key of the object should
-     * be used for both filtering and displayed the selected value in the
-     * textbox.
+     * The currently selected option.
      */
-    optionKey?: OPTION extends Record<string, unknown>
-      ? keyof OPTION
-      : undefined;
-
-    /**
-     * The currently selected option.  If `@options` is an array of strings, provide a string.  If `@options` is an array of objects, pass the entire object.
-     */
-    selected?: OPTION;
+    selected?: string;
   };
   Blocks: {
     default: [
       {
-        option: OPTION;
+        option: string;
         Option: WithBoundArgs<
           typeof OptionComponent,
           | 'index'
@@ -101,13 +89,11 @@ export interface ToucanFormAutocompleteControlComponentSignature<
   Element: HTMLInputElement;
 }
 
-export default class ToucanFormAutocompleteControlComponent<
-  OPTION extends Option
-> extends Component<ToucanFormAutocompleteControlComponentSignature<OPTION>> {
+export default class ToucanFormAutocompleteControlComponent extends Component<ToucanFormAutocompleteControlComponentSignature> {
   @tracked activeIndex: number | null = null;
-  @tracked inputValue: string | undefined;
+  @tracked inputValue = '';
   @tracked isPopoverOpen = false;
-  @tracked filteredOptions: OPTION[] | undefined;
+  @tracked filteredOptions: string[] | null = null;
 
   Chevron = Chevron;
   Option = OptionComponent;
@@ -115,18 +101,13 @@ export default class ToucanFormAutocompleteControlComponent<
 
   constructor(
     owner: unknown,
-    args: ToucanFormAutocompleteControlComponentSignature<OPTION>['Args']
+    args: ToucanFormAutocompleteControlComponentSignature['Args']
   ) {
     super(owner, args);
 
     // We need to set our input tag's value attribute
     // if we have a selected option provided on render
-    let { selected, optionKey } = this.args;
-
-    this.inputValue =
-      typeof selected === 'object' && optionKey
-        ? (selected[optionKey] as string | undefined)
-        : (selected as string | undefined);
+    this.inputValue = this.args.selected ?? '';
   }
 
   velcroMiddleware: VelcroMiddleware[] = [
@@ -231,8 +212,6 @@ export default class ToucanFormAutocompleteControlComponent<
       this.args.options !== undefined
     );
 
-    let { optionKey, onChange } = this.args;
-
     this.closePopover();
 
     assert(
@@ -242,7 +221,7 @@ export default class ToucanFormAutocompleteControlComponent<
 
     // This shouldn't be possible, but to satisfy TS
     if (this.activeIndex === null) {
-      onChange?.(null);
+      this.args.onChange?.(null);
 
       return;
     }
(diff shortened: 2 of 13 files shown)


packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -101,32 +89,25 @@ export interface ToucanFormAutocompleteControlComponentSignature<
...
-export default class ToucanFormAutocompleteControlComponent<
-  OPTION extends Option
-> extends Component<ToucanFormAutocompleteControlComponentSignature<OPTION>> {
+export default class ToucanFormAutocompleteControlComponent extends Component<ToucanFormAutocompleteControlComponentSignature> {
   @tracked activeIndex: number | null = null;
-  @tracked inputValue: string | undefined;
+  @tracked inputValue = '';
   @tracked isPopoverOpen = false;
-  @tracked filteredOptions: OPTION[] | undefined;
+  @tracked filteredOptions: string[] | null = null;
```
clintcs, line 96: Switched to `null` to match `activeIndex`.

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -253,24 +232,17 @@ export default class ToucanFormAutocompleteControlComponent<
...
-    this.args.onChange?.(this.options[this.activeIndex]);
-
-    this.filteredOptions = undefined;
+    this.args.onChange?.(this.options[this.activeIndex] ?? null);
+    this.filteredOptions = null;
   }
 
   @action
-  isEqual(one: number | Option | null, two: number | Option | null) {
+  isEqual(one: unknown, two: unknown) {
```
clintcs, line 240: We don't need to know the incoming type 👍

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -398,33 +365,16 @@ export default class ToucanFormAutocompleteControlComponent<
...
-      filteredOptions = onFilter(value);
-    }
+    this.filteredOptions =
+      this.args.options && this.args.onFilter
+        ? this.args.onFilter(value)
+        : this.args.options
+        ? this.args.options.filter((option) => {
+            return option.toLowerCase().startsWith(value.toLowerCase());
+          })
+        : null;
```
ynotdraw, line 375: Soooo much cleaner!

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -465,48 +415,41 @@ export default class ToucanFormAutocompleteControlComponent<
    *    option.
    */
   @action
-  resetValue(event: Event) {
+  resetEverything(event: Event) {
```
ynotdraw, line 418: I dig the name change 👍

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -465,48 +415,41 @@ export default class ToucanFormAutocompleteControlComponent<
...
+      this.args.onChange?.(null);
 
       return;
     }
 
     if (
-      typeof selected === 'object' &&
-      optionKey &&
-      event.target.value !== selected[optionKey]
+      typeof this.args.selected === 'string' &&
```
ynotdraw, line 439: Is `typeof this.args.selected === 'string'` still needed?

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -465,48 +415,41 @@ export default class ToucanFormAutocompleteControlComponent<
...
+      return;
     }
   }
 
   /**
    * Highlights the entered value of the input when the input is clicked.
    */
   @action
-  selectInput(event: Event) {
+  highlightInputValue(event: Event) {
```
ynotdraw, line 452: Good rename here as well 👍

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
clintcs: Hey! Get out of my draft! 😂

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
clintcs: Nope! Nixed.

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
```diff
@@ -52,39 +48,31 @@ export interface ToucanFormAutocompleteControlComponentSignature<
...
-    onChange?: (option: unknown) => void;
+    onChange?: (selected: string | null) => void;
 
     /**
      * The function called when a user types into the textbox, typically used to write custom filtering logic.
+     *
+     * @return An array of options to be shown in the popover.
      */
-    onFilter?: (input: string) => OPTION[];
+    onFilter?: (value: string) => string[];
```
clintcs, line 58: @ynotdraw Allowing the consumer to specify an array of filtered options means that what's displayed in the popover can diverge totally from the `@options` argument. It could be a totally different set of options—a whole different array—which means the consumer has two ways to specify options and which set is rendered will depend on whether the user has filtered.

ynotdraw (approved):

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
clintcs: I can dig it. After noodling it a bit, the only thing that's a little off to me is that we'd be asking the consumer to do something imperative (add an event listener) for what previously had API support. The consumer would also have to deal with `event.target` and do an assertion on it before grabbing its `value`.

ynotdraw: > After noodling it a bit, the only thing that's a little off to me is that we'd be asking the consumer to do something imperative (add an event listener) for what previously had API support. The consumer would also have to deal with `event.target` and do an assertion on it before grabbing its `value`, eh?

Yeah, good point.  That could be a downside to folks if they use `@onFilter` frequently.

> Another option might be to give the consumer a filtering callback that's called with each iteration of `@options`. Or what, I think, @simonihmig suggested, where `@onFilter` simply calls Array.prototype.filter.

This is true!  We could do as Simon suggested here as the default to have _some_ baked in filtering support.  Then anything more advanced, folks could use the modifier and adjust `@options` possibly?

clintcs: I'm cool with `Array.prototype.filter`. 

What do you have in mind for the advanced case? I ask because, if we don't have a use-case in mind, then we might forgo suggesting it in the documentation as an alternative. Otherwise, if we have a reasonable advanced use-case, then I'll go ahead and document the possibility of the modifier.

clintcs:
Pushed a changeset.

.changeset/smart-mice-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@crowdstrike/ember-toucan-core': minor
+'@crowdstrike/ember-toucan-form': minor
+---
+
+Removed from Autocomplete support for `@options` as an array of objects.
```
clintcs, line 6: I'd expand on this and add examples, but Autocomplete hasn't been released yet. So no need, as far as I can tell, to give consumers backstory, reasoning, or a migration path.

packages/ember-toucan-core/src/components/form/controls/autocomplete.ts
ynotdraw: Awesome, sounds good to me.

> What do you have in mind for the advanced case? I ask because, if we don't have a use-case in mind, then we might forgo suggesting it in the documentation as an alternative.

Maybe we can chat about this this week but punt on it for now?

PR #227, opened 2023-07-24
Add form.Multiselect support to ToucanForm component

ynotdraw:
## 🚀 Description

This PR exposes `form.Multiselect` from `ToucanForm`.

---

## 🔬 How to Test

- Green build
- Visit https://baf6a022.ember-toucan-core.pages.dev/docs/toucan-form/changeset-validation
- Click submit
- Verify an error is displayed for the multi select
- Select an option and then blur the input
- Verify the error is removed

---

## 📸 Images/Videos of Functionality

Screenshot 2023-07-24 at 9 23 40 AM

PR #229, opened 2023-07-24
chore: Multiselect tweaks

clintcs:
### Only accept an array of strings for `@options`

The consumer is currently allowed to pass either an array of strings _or_ an array of objects. @ynotdraw and I talked about it. We think removing support for arrays of objects makes for an overall better component. 

This change does move some work to the consumer. If the consumer's data structure is anything other than an array of strings, the consumer will now have to do a lookup after `@onChange(selected: string[])` is called to find the selected object in the `@options` array. Consumers will also have to map their data to our `string[]` structure, though this is easily done.

Despite this additional work for the consumer, we think that the **disadvantages** of supporting an array of objects make this change worthwhile:

#### The implementation is more complex

Various [conditionals](https://github.com/CrowdStrike/ember-toucan-core/pull/229/files#diff-18271f9cbb7b93358816e2d878f5652b79f186067405c74744e3556eb4beb8e6L371) and [casts](https://github.com/CrowdStrike/ember-toucan-core/pull/229/files#diff-18271f9cbb7b93358816e2d878f5652b79f186067405c74744e3556eb4beb8e6L379) [were added throughout](https://github.com/CrowdStrike/ember-toucan-core/pull/229/files#diff-18271f9cbb7b93358816e2d878f5652b79f186067405c74744e3556eb4beb8e6L581) to support the two data structures.

#### The API is more complex

 A generic (`OPTION`) was [added](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-837203e404c8fe1aa1e4644e395e8232a778a58071bc041170df58b22e60eeebL105) to (I believe) support `Form`. @ynotdraw can give you details if you're curious.

The [`@optionKey`](https://github.com/CrowdStrike/ember-toucan-core/pull/226/files#diff-837203e404c8fe1aa1e4644e395e8232a778a58071bc041170df58b22e60eeebL75) argument was added and required to be passed by the consumer when `@options` is an object.

#### The slope is slippery

What about consumers who want to supply an array of objects where the desired data doesn't reside on the top level of each object in the array? Or maybe the consumer has an array of arrays or a mix of arrays and objects? 

The permutations are endless, and we're bound to encounter all of them given enough consumers. Thus supporting an array of objects isn't just that—it's supporting arbitrary data structures, which is not the business we want to be in.

### Various tweaks and nitpicks

🤷‍♂️

## 🔬 How to Test (Copied from the [original PR](https://github.com/CrowdStrike/ember-toucan-core/pull/227))

1. Green build
1. Visit https://725ae457.ember-toucan-core.pages.dev/docs/components/multiselect
1. Click submit
1. Verify an error is displayed for the multi select
1. Select an option and then blur the input
1. Verify the error is removed

## Related

[Autocomplete tweaks PR](https://github.com/CrowdStrike/ember-toucan-core/pull/226)

### DIFF
--- a/packages/ember-toucan-core/src/components/form/controls/multiselect.ts
+++ b/packages/ember-toucan-core/src/components/form/controls/multiselect.ts
@@ -18,11 +18,7 @@ import Cross from '../../../-private/icons/cross';
 import type { Middleware as VelcroMiddleware } from '@floating-ui/dom';
 import type { WithBoundArgs } from '@glint/template';
 
-export type Option = string | Record<string, unknown> | undefined;
-
-export interface ToucanFormMultiselectControlComponentSignature<
-  OPTION extends Option
-> {
+export interface ToucanFormMultiselectControlComponentSignature {
   Args: {
     /**
      * A CSS class to add to this component's content container.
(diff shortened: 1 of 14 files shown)


packages/ember-toucan-core/src/components/form/controls/multiselect.ts
```diff
@@ -108,13 +94,11 @@ export interface ToucanFormMultiselectControlComponentSignature<
...
-export default class ToucanFormMultiselectControlComponent<
-  OPTION extends Option
-> extends Component<ToucanFormMultiselectControlComponentSignature<OPTION>> {
+export default class ToucanFormMultiselectControlComponent extends Component<ToucanFormMultiselectControlComponentSignature> {
   @tracked activeIndex: number | null = null;
-  @tracked inputValue: string | undefined;
+  @tracked inputValue = '';
   @tracked isPopoverOpen = false;
-  @tracked filteredOptions: OPTION[] | undefined;
+  @tracked filteredOptions: string[] | null = null;
```
clintcs, line 101: Switched to `null` to match `activeIndex`.

ynotdraw (approved):

PR #230, opened 2023-07-25
fix: centre disabled button label

joelamb:
we want button text to be centred, even when the parent item causes the button to be wider than the label text

this is a follow up to addition of flex-grow that allows wrapped buttons to expand to the full parent width in b7472c5851


clintcs:
It's a small change, but consistency is nice. Mind adding some testing steps next time and a screenshot? 🙂

packages/ember-toucan-core/src/components/button.hbs
```diff
@@ -9,7 +9,7 @@
     {{yield to="loading"}}
     <span class="sr-only" data-loading>{{yield}}</span>
   {{else if @isDisabled}}
-    <span class="flex flex-grow items-center gap-x-2">
+    <span class="flex flex-grow items-center justify-centre gap-x-2">
```
ynotdraw, line 12: Should this be:

```suggestion
    <span class="flex flex-grow items-center justify-center gap-x-2">
```

looking at the [Tailwind docs](https://tailwindcss.com/docs/justify-content#center)

joelamb: oops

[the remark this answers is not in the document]
joelamb: I wish :)

joelamb:
typo fixed... interesting that lint rules didn't catch that

ynotdraw:
Looks good to me! Will want to add a `patch` changeset though 👍

clintcs (approved):

ynotdraw (approved):

packages/ember-toucan-core/src/components/button.hbs
nicolechung: This does make me wonder...we have [prettier for tailwind](https://github.com/CrowdStrike/ember-toucan-core/blob/1c1d185bdb49f11cf937cf2797d112aef5983e1b/docs-app/.prettierrc.js#L14) but not any [eslint for tailwind](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/no-custom-classname.md)...should we care about this?

Asking as someone only somewhat familiar with the tailwind linting setup...

PR #240, opened 2023-07-27
chore: make Autocomplete `@noResultsText` required

clintcs:
## 🚀 Description

Makes `@noResultsText` required as a followup to [this discussion](https://github.com/CrowdStrike/ember-toucan-core/pull/238#discussion_r1276486527).

- Makes Autocomplete `@noResultsText` required
- Adds "Optional" descriptors to Autocomplete's arguments documentation
- Add "No results text" to Autocomplete Field documentation
- Moves Multiselect's `@noResultsText` documentation to the top because it's required

## 🔬 How to Test

We're good as long as the build is green.

### DIFF
--- a/docs/components/autocomplete/index.md
+++ b/docs/components/autocomplete/index.md
@@ -3,7 +3,29 @@
 Provides a Toucan-styled autocomplete with filtering.
 If you are building forms, you may be interested in the [AutocompleteField](./autocomplete-field) component instead.
 
+## No results text
+
+Required.
+
+`@noResultsText` is shown when there are no results after filtering. 
+
+```hbs
+<Form::Controls::Autocomplete
+  @noResultsText='No results'
+  @options={{this.options}}
+  @selected={{this.selected}}
+  as |autocomplete|
+>
+  <autocomplete.Option>
+    <!-- The content of each popover list item will be rendered here -->
+    {{autocomplete.option}}
+  </autocomplete.Option>
+</Form::Controls::Autocomplete>
+```
+
 ## Popover z-index
+
+Optional.
 
 A CSS class to add to this component's content container. Commonly used to specify a `z-index`.
 
@@ -12,6 +34,8 @@ A CSS class to add to this component's content container. Commonly used to speci
 ```
 
 ## Options
+
+Optional.
 
 `@options` forms the content of this component. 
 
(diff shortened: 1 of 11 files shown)


clintcs:
Pushed a second commit with the following improvements: 

- Added "Optional" descriptors to Autocomplete's arguments documentation
- Added "No results text" to Autocomplete Field's documentation
- Moved documentation for required arguments to the top of Autocomplete and Multiselect

docs/components/autocomplete/index.md
```diff
@@ -148,19 +182,21 @@ export default class extends Component {
...
-  handleFilter(value) {
+  onFilter(value) {
     return this.options.filter((option) => option === value);
   }
 }
 ```
 
 ## Disabled State
 
+Optional.
```
clintcs, line 198: Unlike the arguments documentation above, the documentation here is for states that reference arguments—so they're very slightly different. @ynotdraw: Should these states have an "Optional" descriptor to match the arguments documentation? Or do you think of them differently? If the former, we might make a note to add "Optional" to other components' state documentation.

docs/components/multiselect-field/index.md
```diff
@@ -198,6 +150,54 @@ Required.
 </Form::Fields::Multiselect>
 ```
 
+## Hint
```
clintcs, line 153: Moved down so the documentation for "No results text" is next to the other required arguments. @ynotdraw: Required-argument documentation at the top across the board? If you like, I can do an audit and then come back around to tweak documentation for the other components.

docs/components/autocomplete/index.md
ynotdraw: I can dig having "Optional" for these like you have in this PR 🎉

docs/components/multiselect-field/index.md
ynotdraw: Sounds great to me. Yeah I think maybe moving the required to the top is likely best as if folks don't provide those arguments they'll run into issues. I think when we initially started we may have documented by personal preference / personal importance, but I think standardizing on maybe required first, then optional makes sense. I appreciate the thinking here!

ynotdraw (approved):

docs/components/multiselect-field/index.md
clintcs: Will do!

docs/components/autocomplete/index.md
clintcs: Great! I'll open a followup.

PR #243, opened 2023-08-01
Upgrade to PNPM 8

clintcs:
## 🚀 Description

Upgrades the repository to [PNPM 8](https://github.com/pnpm/pnpm/releases/tag/v8.0.0). 

- The feature we care most about is probably `auto-install-peers`, which is enabled by default and will remove peer dependency warnings from our installation logs.

- Renovate [opened](https://github.com/CrowdStrike/ember-toucan-core/pull/203) an upgrade PR for us, but it's incomplete. I'll decline it after this PR is merged.

- I've omitted a Changeset. This change doesn't affect consumers—only developers. 

## 🔬 How to Test

Build passed? We're good.


package.json
```diff
@@ -9,14 +9,11 @@
...
-    "pnpm": ">=7.24.2 <7.33.0"
-  },
-  "engines:comment": {
-    "pnpm": "Our lock file version is 6. PNPM 7.24.2 is the first PNPM version to support version 6 lock files. PNPM 7.33.0 adds support for version 6.1 lock files, so using it results in a `pnpm-lock.yaml` diff."
+    "pnpm": ">=8"
   },
   "volta": {
     "node": "18.17.0",
-    "pnpm": "7.32.5"
+    "pnpm": "8.6.11"
```
clintcs, line 16: @ynotdraw: I saw some discussion elsewhere about needing to upgrade Embroider to unlock the PNPM 8 upgrade. Am I missing something?

.github/actions/pnpm/action.yml
```diff
@@ -5,7 +5,7 @@ runs:
   steps:
     - uses: pnpm/action-setup@v2.4.0
       with:
-        version: 7.32.5
+        version: 8.6.11
```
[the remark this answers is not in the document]
ynotdraw: I don't have an opinion here, but was curious what you thought - would it make sense for CI to do `version: 8` instead, so that it always uses the latest version of `pnpm`?  Do we care if CI is on a different version than what we may be running locally?  We were previously on "use latest" like this and caught a bug with `pnpm` that got resolved rather quickly.

There could be some downsides to this though.  I don't know if it's totally worth it, but thought I'd get your opinion.  I can't remember if renovate/dependabot will auto-upgrade this version to match what is in the volta / engines config.

docs-app/package.json
```diff
@@ -114,7 +113,6 @@
     "prettier-plugin-tailwindcss": "^0.4.0",
     "qunit": "^2.19.1",
     "qunit-dom": "^2.0.0",
-    "rsvp": "^4.8.5",
```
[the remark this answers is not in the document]
ynotdraw: Not that I remember! This was the main one I recall.

package.json
ynotdraw: Not that I recall!  If the build and local dev is happy, I'm happy!

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
```
ynotdraw, line 1: One thing I'm curious about and am having a hard time finding after ~5mins of googling is: if we completely delete the `pnpm-lock` file here and re-reun `pnpm i` will this be updated to the latest version (6.1)?  Do we care about that? I'm not sure so thought I'd ask!

ynotdraw (approved):

pnpm-lock.yaml
clintcs: Thanks for asking! I was wondering the same thing. Deleting `pnpm-lock.yaml` and running `pnpm i` creates a lockfile diff for some reason but doesn't bump the lockfile version to 6.1.

From PNPM's release notes, it looks like they bumped the lockfile to 6.1 with PNPM [8.6.0](https://github.com/pnpm/pnpm/releases/tag/v8.6.0) but then downgraded the lockfile back down to 6.0 with PNPM [8.6.2](https://github.com/pnpm/pnpm/releases/tag/v8.6.2). Perhaps that explains why we're not getting a 6.1 lockfile.

.github/actions/pnpm/action.yml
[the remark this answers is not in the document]
clintcs: No opinion here either 🤷‍♂️

pnpm-lock.yaml
ynotdraw: Ah, interesting! Thanks for digging in.

PR #244, opened 2023-08-01
Add "Select all" functionality to multiselect

ynotdraw:
## 🚀 Description

Adds "Select all" functionality to multiselect.

---

## 🔬 How to Test

- Visit https://8b4676f3.ember-toucan-core.pages.dev/docs/components/multiselect
- For this section, use the bottom multiselect ( "With Select All" )

**Select all states**
- Click the multiselect
- Click "Select all"
  - Verify the checkbox is checked
  - Verify all items in the popover are also checked
- Click "Orange"
  - Verify "Orange" is removed from the list
  - Verify the "Select all" checkbox is now in the indeterminate state
- Click "Select all" again
  - Verify all items are selected again and "Orange", in particular, is re-added
- Click "Select all" again
  - Verify all items are unselected
  - Verify the "Select all" checkbox is unchecked

**Filtering**
- Click the multiselect
- Type into the multiselect to trigger filtering.  For example type "b" to filter down to "Blue"
- Verify "Select all" is no longer visible
- Select "Blue"
- Upon selecting, verify the filter clears and the "Select all" checkbox is now in the indeterminate state

Feel free to play more with this functionality!
---

**Select all option**

Screenshot 2023-08-02 at 8 17 05 AM

**Indeterminate state**

Screenshot 2023-08-02 at 8 17 57 AM

**Select all not visible when filtering**

Screenshot 2023-08-02 at 8 18 20 AM

**With a screenreader**

https://github.com/CrowdStrike/ember-toucan-core/assets/8069555/1df36b41-9885-47ce-8b10-817406755158

### DIFF
--- a/docs/components/multiselect/demo/base-demo.md
+++ b/docs/components/multiselect/demo/base-demo.md
@@ -2,7 +2,7 @@
 <div class='flex flex-col gap-4 w-96'>
   <Form::Controls::Multiselect
     @contentClass='z-10'
-    @noResultsText="No results"
+    @noResultsText='No results'
     @onChange={{this.onChange}}
     @options={{this.options}}
     @selected={{this.selected}}
@@ -24,7 +24,7 @@
 
   <Form::Controls::Multiselect
     @contentClass='z-10'
-    @noResultsText="No results"
+    @noResultsText='No results'
     @onChange={{this.onChange2}}
     @options={{this.options2}}
     @selected={{this.selected2}}
@@ -46,12 +46,35 @@
 
   <Form::Controls::Multiselect
     @contentClass='z-10'
-    @noResultsText="No results"
+    @noResultsText='No results'
     @onChange={{this.onChange3}}
     @onFilter={{this.onFilter}}
     @options={{this.options}}
     @selected={{this.selected3}}
     placeholder='Colors w/ Filtering'
+  >
+    <:chip as |chip|>
+      <chip.Chip>
+        {{chip.option}}
+        <chip.Remove @label={{(concat 'Remove' ' ' chip.option)}} />
+      </chip.Chip>
+    </:chip>
+
+    <:default as |multiselect|>
+      <multiselect.Option>
+        {{multiselect.option}}
+      </multiselect.Option>
+    </:default>
+  </Form::Controls::Multiselect>
+
+  <Form::Controls::Multiselect
+    @contentClass='z-10'
+    @noResultsText='No results'
+    @onChange={{this.onChange4}}
+    @options={{this.options}}
+    @selected={{this.selected4}}
+    @selectAllText='Select all'
+    placeholder='Colors w/ Select All'
   >
     <:chip as |chip|>
       <chip.Chip>
@@ -78,6 +101,7 @@ export default class extends Component {
   @tracked selected;
   @tracked selected2;
   @tracked selected3;
+  @tracked selected4;
 
   options = [
     {
@@ -145,6 +169,12 @@ export default class extends Component {
   @action
   onChange3(option) {
     this.selected3 = option;
+    console.log(option);
+  }
+
+  @action
+  onChange4(option) {
+    this.selected4 = option;
     console.log(option);
   }
 
--- a/packages/ember-toucan-core/src/components/form/controls/multiselect.ts
+++ b/packages/ember-toucan-core/src/components/form/controls/multiselect.ts
@@ -13,6 +13,7 @@ import OptionComponent, {
   selector as optionComponentSelector,
 } from '../../../-private/components/form/controls/multiselect/option';
 import RemoveComponent from '../../../-private/components/form/controls/multiselect/remove';
+import SelectAllComponent from '../../../-private/components/form/controls/multiselect/select-all';
 import Chevron from '../../../-private/icons/chevron';
 import Cross from '../../../-private/icons/cross';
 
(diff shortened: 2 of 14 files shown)


packages/ember-toucan-core/src/components/form/controls/multiselect.ts
```diff
@@ -176,6 +198,50 @@ export default class ToucanFormMultiselectControlComponent extends Component<Tou
...
+   * Helps us determine if the "Select all" option checkbox
+   * should be checked or not.
+   */
+  get isSelectAllChecked() {
+    // We need to use this comparison method when comparing arrays
+    // as Ember's `isEqual` is only for objects.
+    return (
+      JSON.stringify(this.selected?.sort()) ===
+      JSON.stringify(this.options?.sort())
+    );
```
ynotdraw, line 228: Open to some code golf here, but we can't use [Ember's `isEqual`](https://api.emberjs.com/ember/release/functions/@ember%2Futils/isEqual) because it doesn't play well with comparing arrays.

> Due to the expense of array comparisons, collections will never be equal to each other even if each of their items are equal to each other.

Sorting `selected` and `options` and then doing a JSON stringify comparison is the quickest way I know to do this.  Since we know there won't be duplicates in the arrays (since our business logic doesn't allow for that - re-selecting an already selected item removes it) I think this is probably safe, but I'm also open to other suggestions if anyone feels strongly about improving this.

test-app/tests/integration/components/multiselect-test.gts
```diff
@@ -1501,6 +1502,134 @@ module('Integration | Component | Multiselect', function (hooks) {
...
+
+    await click('[data-multiselect]');
+
+    assert.dom('[data-multiselect-select-all-option]').exists();
+    assert.dom('[data-multiselect-select-all-option]').hasText('Select all');
+
+    assert.dom('[data-multiselect-select-all-checkbox]').exists();
+  });
+
+  test('it handles all possible state changes when the `Select all` option is interacted with', async function (assert) {
```
ynotdraw, line 1534: This test is kind of big, but for this case it was easier to test the entire functionality rather than breaking it into smaller tests due to requiring particular states of the select all checkbox.

test-app/tests/integration/components/multiselect-test.gts
```diff
@@ -1501,6 +1502,134 @@ module('Integration | Component | Multiselect', function (hooks) {
...
+    assert.expect(22);
+
+    class State {
+      @tracked selected: string[] = [];
+    }
+
+    let state = new State();
```
ynotdraw, line 1541: We have to use `State` here because we need `@selected` passed into `Multiselect` to be `@tracked`; otherwise, our value will not update when the `handleChange` sets the new value. So a bit more complicated than our other tests.

test-app/tests/integration/components/multiselect-test.gts
```diff
@@ -1501,6 +1502,134 @@ module('Integration | Component | Multiselect', function (hooks) {
...
+        <:default as |multiselect|>
+          <multiselect.Option>{{multiselect.option}}</multiselect.Option>
+        </:default>
+      </Multiselect>
+    </template>);
+
+    // Open the popover
+    await click('[data-multiselect]');
+
+    // Verify default state of `Select all` is not checked
```
ynotdraw, line 1573: Tried being overly verbose here on the expectations for each step along the way

packages/ember-toucan-core/src/components/form/controls/multiselect.ts
ynotdraw: - [ ] Add this comment in code!

packages/ember-toucan-core/src/components/form/controls/multiselect.ts
```diff
@@ -176,6 +198,50 @@ export default class ToucanFormMultiselectControlComponent extends Component<Tou
...
+   * Helps us determine if the "Select all" option checkbox
+   * should be indeterminate or not.
+   */
+  get isSelectAllIndeterminate() {
+    return (
+      this.selected?.length > 0 &&
+      // We need to use this comparison method when comparing arrays
+      // as Ember's `isEqual` is only for objects.
+      JSON.stringify(this.selected?.sort()) !==
+        JSON.stringify(this.options?.sort())
```
ynotdraw, line 241: - [ ] Add comment here about reasoning

packages/ember-toucan-core/src/components/form/controls/multiselect.ts
clintcs: Seems like the simplest way. IMO not worth bringing in Lodash just for this.

test-app/tests/integration/components/multiselect-test.gts
clintcs: Made it very easy for me to follow along 👍

clintcs (approved):

docs/components/multiselect/demo/base-demo.md
```diff
@@ -66,6 +66,29 @@
...
   </Form::Controls::Multiselect>
+
+  <Form::Controls::Multiselect
+    @contentClass='z-10'
+    @noResultsText='No results'
+    @onChange={{this.onChange4}}
+    @options={{this.options}}
+    @selected={{this.selected4}}
+    @selectAllText='Select all'
+    placeholder='With Select All'
```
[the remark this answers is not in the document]
ynotdraw: I dig it - pushed!

Issue #210, opened 2023-07-11
Get this repo working with embroider v3+


ynotdraw:
As part of https://github.com/CrowdStrike/ember-toucan-core/pull/200, we are disabling embroider-optimized tests to unblock our work.  We started getting the following errors with the embroider-optimized tests with that PR:

```bash
test-app:test: [1]   - message: ../node_modules/.pnpm/file+packages+ember-toucan-core_@babel+core@7.20.1_chhstik6enqihwk47y4lzcbm7a/node_modules/@crowdstrike/ember-toucan-core/dist/components/form/controls/combobox.js: Module not found: Error: @crowdstrike/ember-toucan-core is trying to import from #embroider_compat but that is not one of its explicit dependencies
```

The suggestion was to attempt to upgrade `pnpm` and `embroider` dependencies to the latest as done in https://github.com/CrowdStrike/ember-toucan-core/pull/205; however, we still have a few issues with those changes.  In particular:

```bash
test-app:test: [1] $TMPDIR/embroider/c6249b/node_modules/.pnpm/file+packages+ember-toucan-core_@babel+core@7.20.12_@crowdstrike+ember-toucan-styles@2.0.2_@e_tm5lpotvrba5vzg34cuvh2fqxm/node_modules/@crowdstrike/ember-toucan-core/dist/components/form/controls/combobox.js/combobox.js: Unsafe dynamic component: this.Option in $TMPDIR/embroider/c6249b/node_modules/.pnpm/file+packages+ember-toucan-core_@babel+core@7.20.12_@crowdstrike+ember-toucan-styles@2.0.2_@e_tm5lpotvrba5vzg34cuvh2fqxm/node_modules/@crowdstrike/ember-toucan-core/dist/components/form/controls/combobox.js
```

This is because we removed the `ensure-safe-component` helper as part of that PR.

When adding back the `ensure-safe-component` helper, we get a new error from https://github.com/CrowdStrike/ember-toucan-core/pull/206.  [You can view the full action log here](https://github.com/CrowdStrike/ember-toucan-core/actions/runs/5508111169/jobs/10038990625)

```bash
test-app:test: [1]   - broccoliBuilderErrorStack: ModuleNotFoundError: Module not found: Error: Can't resolve './helpers/ensure-safe-component' in '/home/runner/work/ember-toucan-core/ember-toucan-core/test-app/node_modules/.embroider/rewritten-app'
```

[...]

PR #246, opened 2023-08-03
Convert addons to template tag and gts

ynotdraw:
## 🚀 Description

Converts `toucan-core` and `toucan-form` to use the `<template>` tag and `gts` file extension.  A couple of things worth mentioning:

### Massive PR

Unfortunately this PR is massive; however, not much code was actually changed.  It was mostly creating a `gts` file, then copying the `hbs` file into the `<template>` and then adding any helper imports needed.  While I was in the code I did a few things like added missing JSdocs or renamed things, but everything else stayed put.

### Tailwind CSS Autocomplete

**NOTE:** This only applies to you if you use VSCode.

In VS code, it does not recognize how to format `gjs/gts` class names by default.  Due to that, you need to help the extension out by telling it which language glimmer-ts is.  We map it to "html".

Go to the settings for the TailwindCSS extension and add the following:

Screenshot_2023-08-03_at_8 57 01_AM

### `prettier-plugin-tailwindcss`
The downside with `gjs/gts` is that it appears our `prettier-plugin-tailwindcss` rule no longer works.  This rule would auto-sort the class names for us based on Tailwind's rules.  I believe we may need to update the prettier glimmer plugin, but I asked on the Ember Discord for clarification.  I could also be missing something.  At the time I was thinking maybe we needed something like above to help tell the underlying library how to handle these files, but since it's prettier rather than an extension, we don't appear to have the correct hooks for that configuration.

I suppose the good news would be that once the underlying issue is resolved, it should all just work in a future `prettier` release.

https://discord.com/channels/480462759797063690/486516209873846292/1136646656977289236

---

## 🔬 How to Test

- Green build
- Generally poking around https://51b75e02.ember-toucan-core.pages.dev/ is probably wise too

---

N/A


docs-app/package.json
```diff
@@ -36,7 +36,7 @@
     "@ember/string": "^3.0.1",
     "@ember/test-helpers": "^3.1.0",
     "@embroider/compat": "^2.0.0",
-    "@embroider/core": "^2.0.0",
+    "@embroider/core": "2.0.2",
```
ynotdraw, line 39: Had to pin this for now due to an error I was seeing.  The task to upgrade us to the latest embroider stuff is at https://github.com/CrowdStrike/ember-toucan-core/issues/210

The error was:

```bash
docs-app:build: $TMPDIR/embroider/6cc172/node_modules/.pnpm/file+packages+ember-toucan-core_@babel+core@7.20.12_@crowdstrike+ember-toucan-styles@2.0.1_@e_y4sbboigrbv5roskhqf7pymkbu/node_modules/@crowdstrike/ember-toucan-core/dist/components/form/fields/textarea.js: Scope objects for `precompileTemplate` may only contain direct references to in-scope values, e.g. { Field } or { Field: Field }
docs-app:build:   115 |   `, {
docs-app:build:   116 |   strictMode: true,
docs-app:build: > 117 |   scope: () => ({
docs-app:build:       |   ^
docs-app:build:   118 |     Field: ToucanFormFieldComponent,
docs-app:build:   119 |     hash,
docs-app:build:   120 |     field,
```

docs-app/package.json
```diff
@@ -108,7 +108,7 @@
     "postcss": "^8.4.17",
     "postcss-import": "^15.0.0",
     "postcss-loader": "^7.0.1",
-    "prettier": "^2.8.3",
+    "prettier": "^3.0.1",
```
ynotdraw, line 111: Needed to update prettier so that the prettier template plugin would work properly

packages/ember-toucan-core/.prettierrc.cjs
ynotdraw, line 1: Removed this one as we have one in the root.  This was likely left over from when we moved from the single-package to the multiple-package format.

packages/ember-toucan-core/package.json
```diff
@@ -72,6 +73,8 @@
     "@types/ember__debug": "^4.0.0",
     "@types/ember__engine": "^4.0.0",
     "@types/ember__error": "^4.0.0",
+    "@types/ember__helper": "^4.0.2",
+    "@types/ember__modifier": "^4.0.3",
```
ynotdraw, line 77: With `gjs/gts` we need to import helpers like `on`, `fn`, and `hash` explicitly.  By adding these dependencies, it allows us to do so (see below!)

packages/ember-toucan-core/src/-private/components/form/controls/autocomplete/option.gts
```diff
@@ -1,9 +1,10 @@
 import Component from '@glimmer/component';
+import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 
 import Check from '../../../../../-private/icons/check';
 
-interface ToucanFormAutocompleteOptionControlComponentSignature {
+interface ToucanCoreAutocompleteOptionComponentSignature {
```
ynotdraw, line 7: For these `-private` components, I updated the prefix to `ToucanCore*`

packages/ember-toucan-core/src/components/button.gts
```diff
@@ -76,9 +77,9 @@ export default class Button extends Component<ButtonSignature> {
 
     assert(
       `Invalid variant for Button: '${variant}' (allowed values: [${VALID_VARIANTS.join(
-        ', '
+        ', ',
```
ynotdraw, line 80: I believe this change was due to the latest version of prettier where trailing commas are now the default.

.changeset/good-rivers-train.md
ynotdraw, line 1: I figured I'd do a patch upgrade _just in case_ things would break consumers for some strange reason 🤞

.prettierrc.js
ynotdraw, line 1: For consistency I took these settings from `ember-headless-form`.

docs-app/package.json
clintcs: Mind adding a comment somewhere? One possibility:

```
`"devDependencyComments": {
  "@embroider/core": "[your comment]"
},`

File: packages/ember-toucan-core/package.json (hunk 3 above)
Comment (clintcs) [reply packages/ember-toucan-core/package.json:77]: Being able to trace helpers back to `import` statements is perhaps my favorite part of `gjs/gts`.

File: packages/ember-toucan-core/rollup.config.mjs [hunk 6]
```diff
@@ -26,11 +27,15 @@ export default {
     // not everything in publicEntrypoints necessarily needs to go here.
     addon.appReexports(['components/**/*.js']),
 
+    // compile <template> tag into plain JS
+    glimmerTemplateTag({ preprocessOnly: true }),
```
clintcs, line 31: Off the top of your head, any idea what is compiles to when there's a backing class? A `template` class method maybe?

clintcs (approved):

packages/ember-toucan-core/package.json
ynotdraw: I 100% agree 🎉

packages/ember-toucan-core/rollup.config.mjs
ynotdraw: I'm honestly not sure but Preston would know!

PR #248, opened 2023-08-07
Add Autocomplete test helpers

clintcs:
## 🚀 Description

- Adds Autocomplete test helpers
- Reworks the Autocomplete tests to use the new test helpers

---

## 🔬 How to Test

We're good if the build is green.

### DIFF
--- a/packages/ember-toucan-core/src/-private/components/form/controls/autocomplete/option.gts
+++ b/packages/ember-toucan-core/src/-private/components/form/controls/autocomplete/option.gts
@@ -96,12 +96,12 @@ export default class ToucanCoreAutocompleteOptionComponent extends Component<Tou
   <template>
     {{! template-lint-disable require-presentational-children }}
     <li
+      aria-current={{if @isActive "true" "false"}}
       aria-selected={{if @isSelected "true" "false"}}
       class="my-0 flex cursor-default items-center gap-2 px-2 py-2 leading-4
         {{this.styles}}
         {{this.className}}
         "
-      data-active={{if @isActive "true" "false"}}
       id="{{@popoverId}}-{{@index}}"
       role="option"
       {{on "click" this.onClick}}
--- a/packages/ember-toucan-core/src/components/form/controls/autocomplete.gts
+++ b/packages/ember-toucan-core/src/components/form/controls/autocomplete.gts
@@ -468,123 +468,125 @@ export default class ToucanFormAutocompleteControlComponent extends Component<To
       @placement="bottom-start"
       as |velcro|
     >
-      <div class="relative flex w-full items-center">
-        {{! template-lint-disable no-redundant-role  }}
-        <input
-          aria-activedescendant={{this.activeDescendant}}
-          aria-autocomplete="list"
-          aria-controls={{this.popoverId}}
-          aria-expanded={{this.isPopoverOpen}}
-          aria-haspopup="listbox"
-          autocapitalize="none"
-          autocomplete="off"
-          autocorrect="off"
-          class="focus:outline-none w-full rounded-sm border-none py-1 pl-2 pr-6 transition-shadow
-            {{this.styles}}"
-          disabled={{@isDisabled}}
-          readonly={{@isReadOnly}}
-          role="combobox"
-          spellcheck="false"
-          type="text"
-          value={{this.inputValue}}
-          {{on
-            "blur"
-            (if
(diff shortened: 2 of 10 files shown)


packages/ember-toucan-core/src/test-support/components/autocomplete.ts
```diff
@@ -0,0 +1,33 @@
+import { PageObject } from 'fractal-page-object';
+
+export class AutocompletePageObject extends PageObject {
+  get active() {
+    return this.element
```
clintcs, line 5: We spread `attributes` on the `<input />`, which seems right to me. However, it means we need to first get ahold of the parent element (`#autocomplete`) before we can select children.

packages/ember-toucan-core/src/-private/components/form/controls/autocomplete/option.gts
```diff
@@ -96,12 +96,12 @@ export default class ToucanCoreAutocompleteOptionComponent extends Component<Tou
   <template>
     {{! template-lint-disable require-presentational-children }}
     <li
+      aria-current={{if @isActive "true" "false"}}
```
clintcs, line 99: image

clintcs: Replaces `data-active` (below).

ynotdraw: Oh very nice!

packages/ember-toucan-core/src/components/form/controls/autocomplete.gts
```diff
@@ -468,123 +468,125 @@ export default class ToucanFormAutocompleteControlComponent extends Component<To
...
-                    this.Option
-                    isActive=(this.isEqual index this.activeIndex)
-                    isDisabled=@isDisabled
-                    isSelected=(this.isEqual @selected option)
-                    isReadOnly=@isReadOnly
-                    onClick=(fn this.onChange index)
-                    onMouseover=(fn this.onOptionMouseover index)
-                    popoverId=this.popoverId
-                    index=index
+      <div id="autocomplete">
```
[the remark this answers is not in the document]
ynotdraw: Mostly just wondering, but any reason to use `id` over ` data-` attribute? `closest` should work with those as well, right?

packages/ember-toucan-core/src/test-support/components/button.ts
```diff
@@ -2,7 +2,7 @@ import { click } from '@ember/test-helpers';
 
 import { PageObject } from 'fractal-page-object';
 
-export class Button extends PageObject {
+export class ButtonPageObject extends PageObject {
```
ynotdraw, line 5: Due to an export name change, should we consider a breaking-change changeset?  I'd be okay making it a `patch`, as I'm not sure anyone is using this directly.  As long as it's documented in the release notes 👍

ynotdraw (commented):
Shall we add a changeset as new test-helpers will be exposed?

packages/ember-toucan-core/src/test-support/components/button.ts
clintcs: I'm cool with that. 

Though I tend to wait to add a changeset until after PR review in case there's some discussion. I'll make sure to add one after approval.

packages/ember-toucan-core/src/components/form/controls/autocomplete.gts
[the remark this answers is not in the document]
clintcs: Oh, for sure. I went back and forth on it. I'll spare you my reasoning and switch it to `data-autocomplete`.

As an aside, if we ever rebuild these components, I'd favor a a slightly different form (`data-test="autocomplete"`) throughout so it's clear what these selectors are for when perusing the templates.

ynotdraw (approved):

packages/ember-toucan-core/src/components/form/controls/autocomplete.gts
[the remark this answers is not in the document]
clintcs: > I'll spare you my reasoning and switch it to data-autocomplete.

Done

packages/ember-toucan-core/src/test-support/components/button.ts
clintcs: Done.

PR #257, opened 2023-08-16
chore: remove Button's test helpers

clintcs:
## 🚀 Description

I've been going through our components and adding test helpers so consumers don't have to select elements themselves, thereby digging into our implementation. It occurred to me while thinking about what needs test helpers and what doesn't that Button may not need any. 

Because Button is more or less composed of one element (`<button />`) the consumer can spread an ID or data attribute onto Button and use that ID or data attribute to select that element. He can then use a combination of QUnit DOM and Ember Test Helpers to assert what needs asserting. No need, for example, to provide an abstraction for checking whether Button is disabled or for clicking it.

---

## 🔬 How to Test

If the build is green, we're good.


clintcs:
@ynotdraw: Before I take this out of draft, add a description, and send it out for review, I thought I'd see how you feel. I'm working on adding test helpers to components that need them, but I'm wondering—does Button need them?

test-app/tests/integration/components/button-test.gts
```diff
@@ -55,8 +49,6 @@ module('Integration | Component | button', function (hooks) {
       </Button>
     </template>);
 
-    assert.false(buttonPageObject.isLoading);
```
clintcs, line 58: This is the only test helper I was able to see an argument for. Is it something we need to test though? Or does the `[data-test-loading]` assertion below sufficiently cover it?

ynotdraw: I feel as though the assertion below is sufficient enough probably

ynotdraw (approved):

ynotdraw:
> Before I take this out of draft, add a description, and send it out for review, I thought I'd see how you feel. I'm working on adding test helpers to components that need them, but I'm wondering: does Button need them?

Good question.  I'm kind of thinking maybe these aren't super helpful?  All of this stuff can be accomplished already with the built-in ember qunit assertions and/or with the built-in ember test helpers.  I think components like our autocomplete and multi select are complicated enough that they warrant helpers; however, I'm not sure I feel the same with `Button`.

A bit tangential to this, but I also wonder if we should make `Button` more dumb eventually and not have those named blocks with styling.  I feel as though they are implementation details that downstream/other consumers may not care about or want.  And those states are something they can write themselves, rather than it being baked into the core `Button`.  Nothing we need to do now, of course, but maybe something in the future to think about to make this less opinionated and force strong opinions _upstream_

PR #300, opened 2024-01-08
Increase renovate's prConcurrentLimit

simonihmig:
By default, we only allow 10 concurrent PRs. If some PRs are red and not trivial to fix, and new updates are only created once a week (see `"schedule": ["after 9pm on sunday"]`), then this can easily lead to a larger backlog, see https://github.com/CrowdStrike/ember-toucan-core/issues/18


ynotdraw (approved):
Great idea

PR #304, opened 2024-01-11
Add event.preventDefault() when clicking on disabled button

davideferre:
## 🚀 Description

When a type="submit" button has @isDisabled set, it does not prevent submitting the form, because the disabled state (for a11y reasons) does not add the disabled attribute.
Adding an `event.preventDefault()` prevent to send the submit event.

---

## 🔬 How to Test
Create a form with a submit action and add a `<Button @isDisabled={{true}} type="submit">button</Button>` inside the form and verify that the submit action is not fired when clicking the button.

### DIFF
--- a/.changeset/afraid-vans-attend.md
+++ b/.changeset/afraid-vans-attend.md
@@ -1,0 +1,5 @@
+---
+"@crowdstrike/ember-toucan-core": patch
+---
+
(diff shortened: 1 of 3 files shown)


simonihmig (approved):
Thank you!

simonihmig:
I added a changeset for releasing this as a new patch version, fyi

ynotdraw (approved):
Thanks!

Issue #446, opened 2024-11-25
Textarea value gets decoupled from what's displayed when updating @value after typing in textarea


kmillns:
When using `Form::Controls::Textarea` and you type in the textarea the `value` and `defaultValue` of the underlying textarea get out of sync with each other, so further updates to `@value` update the `defaultValue` but the `value` isn't updated and the textarea doesn't visually update.

This appears to be because the `@value` set up like `<textarea>{{@value}}</textarea>` which is technically setting the `defaultValue` rather than setting the `value` property like `<textarea value={{@value}}>`

[Relevant snippet from MDN](url)

> The <textarea> element specifies its content differently in HTML and JavaScript contexts:
> - In HTML, the initial content of a <textarea> is specified between its opening and closing tags, not as a value attribute.
> - In JavaScript, <textarea> elements have a [value](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement/value) property that can be used to get or set the current content, and [defaultValue](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement/defaultValue) to get and set its initial value (equivalent to accessing the HTML element's text content).

Example of the problem

```
export default class ExampleComponent extends Component {
  @tracked fooBar = 'foobar';

  @action
  updateFooBar(value, e) {
    console.log({ e, value });
    this.fooBar = value;
  }

  @action
  repeatFooBar() {
    console.log('repeat');
    this.fooBar = this.fooBar + 'fooBar';
  }
}
```

```
  <span>{{this.fooBar}}</span>
  <Form::Controls::Textarea
    @name={{this.fooBar}}
    class="font-monospace type-md"
    @label="foobar"
    @value={{this.fooBar}}
    @onChange={{this.updateFooBar}}
  />
  <button type="button" {{on "click" this.repeatFooBar}}>Repeat</button>
```

Steps:
1. Click "Repeat" - this correctly appends to the textarea contents
2. Type in the textarea - this correctly updates
3. Click "Repeat" again - note that it no longer updates the textarea contents
4. Inspect the textarea and check `.value` and `.defaultValue`

![Screenshot 2024-11-20 at 5 11 07 PM](https://github.com/user-attachments/assets/d3ae9b54-d716-4e8e-9d6f-880706578ccd)
[...]

PR #447, opened 2024-11-25
fix: set value attribute on textarea instead of in content to ensure value and defaultValue stay in sync with changes

kmillns:
## 🚀 Description

Fixes issue #446 where the underlying `<textarea>` used in `Form::Controls::Textarea` gets out of sync with `@value` changes after typing in the `<textarea>`.

Also adds tests for both current working functionality of updating value before typing in the textarea and the previously broken case of typing in the textarea and then attempting an update of the value from outside the component.

---

## 🔬 How to Test

Added test cases should cover the functionality changing.

There's no external impact on existing examples to show.

Reversing the change in `textarea.gts` should result in one failed test case for the issue described in #446

### DIFF
--- a/.changeset/spotty-lamps-shake.md
+++ b/.changeset/spotty-lamps-shake.md
@@ -1,0 +1,5 @@
+---
+'@crowdstrike/ember-toucan-core': patch
+---
+
+Update Textarea to ensure underlying textarea value stays in sync with both typing and @value changes
--- a/.changeset/spotty-lamps-shake.md
+++ b/.changeset/spotty-lamps-shake.md
@@ -2,4 +2,4 @@
 '@crowdstrike/ember-toucan-core': patch
 ---
 
-Update Textarea to ensure underlying textarea value stays in sync with both typing and @value changes
+Update Textarea to ensure underlying <textarea> value stays in sync with both typing and @value changes
--- a/packages/ember-toucan-core/src/components/form/controls/textarea.gts
+++ b/packages/ember-toucan-core/src/components/form/controls/textarea.gts
@@ -80,7 +80,8 @@ export default class ToucanFormTextareaControlComponent extends Component<Toucan
         readonly={{@isReadOnly}}
         ...attributes
         {{on "input" this.handleInput}}
-      >{{@value}}</textarea>
+        value={{@value}}
+      ></textarea>
     </div>
   </template>
 }
--- a/test-app/tests/integration/components/textarea-test.gts
+++ b/test-app/tests/integration/components/textarea-test.gts
@@ -1,9 +1,11 @@
 /* eslint-disable no-undef -- Until https://github.com/ember-cli/eslint-plugin-ember/issues/1747 is resolved... */
-import { fillIn, render } from '@ember/test-helpers';
+import { fillIn, render, settled } from '@ember/test-helpers';
 import { module, test } from 'qunit';
 
 import TextareaControl from '@crowdstrike/ember-toucan-core/components/form/controls/textarea';
 import { setupRenderingTest } from 'test-app/tests/helpers';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 
 module('Integration | Component | Textarea', function (hooks) {
   setupRenderingTest(hooks);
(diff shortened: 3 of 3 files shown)


nicolechung:
@kmillns  these changes look correct to me although I am unable to reproduce the original bug in storybook at: https://glide-core.crowdstrike-ux.workers.dev/?path=/docs/textarea--overview

Screenshot 2024-11-26 at 1 53 20 PM

.changeset/spotty-lamps-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@crowdstrike/ember-toucan-core': patch
+---
+
+Update Textarea to ensure underlying textarea value stays in sync with both typing and @value changes
```
nicolechung, line 5: Suggestion, nit, non-blocking:

```suggestion
Update Textarea to ensure underlying <textarea> value stays in sync with both typing and @value changes
```

test-app/tests/integration/components/textarea-test.gts
```diff
@@ -78,6 +80,60 @@ module('Integration | Component | Textarea', function (hooks) {
     assert.dom('[data-textarea]').hasValue('tony');
   });
 
+  test('it keeps the value in sync with external changes to `@value`', async function (assert) {
```
nicolechung, line 83:
```suggestion
test('it keeps the <textarea> value in sync with external changes to `@value`', async function (assert) {
```

test-app/tests/integration/components/textarea-test.gts
```diff
@@ -78,6 +80,60 @@ module('Integration | Component | Textarea', function (hooks) {
...
+    assert.dom('[data-textarea]').hasValue('initial');
+
+    testContext.testValue = 'updated';
+
+    await settled();
+
+    assert.dom('[data-textarea]').hasValue('updated');
+  });
+
+  test('it keeps the value in sync with external changes to `@value` after the value is changed', async function (assert) {
```
nicolechung, line 105:
```suggestion
  test('it keeps the <textarea> value in sync with external changes to `@value` after the value is changed', async function (assert) { 
  ```

**Review by nicolechung — APPROVED** [review]
Thank you for making this fix 🚀

**Comment by nicolechung** [comment]
Screenshot 2024-11-26 at 1 59 06 PM

@kmillns  I'm a little confused as I am unable to re-create the error problem over here: https://ember-toucan-core.pages.dev/docs/components/textarea

But your code and test look correct to me, so I've approved.

**Comment by kmillns** [comment]
> Screenshot 2024-11-26 at 1 59 06 PM
> @kmillns I'm a little confused as I am unable to re-create the error problem over here: https://ember-toucan-core.pages.dev/docs/components/textarea
> 
> But your code and test look correct to me, so I've approved.

Yeah, I don't think it's something that would normally happen with just typing in the `<textarea>` but instead is the combination of typing, which causes the underlying `<textarea>` to use `.value` instead of `.defaultValue` and then triggering a change in the `@value`, which is essentially just changing the `.defaultValue`.

These screenshots show how `.value` and `.defaultValue` (initially both `''` in the demo) get out of sync after typing in the `<textarea>`, and the `@value` being inside only updates the `.defaultValue` so it will only appear out of sync if trying to update after typing.
Screenshot 2024-11-26 at 3 35 38 PM
Screenshot 2024-11-26 at 3 35 54 PM
Screenshot 2024-11-26 at 3 36 09 PM

**Review by ynotdraw — APPROVED** [review]
Looks good to me. Let's see if @simonihmig remembers about the value change above to make sure we don't have a potential issue before merging! Thanks!

**Comment by ynotdraw** [comment]
@kmillns I kicked off CI and looks like a few linting/TS errors

**Comment by kmillns** [comment]
Everything looks green except cloudflare publish which I think is unrelated to this PR:

`Error: Input required and not supplied: apiToken`

**Comment by ynotdraw** [comment]
> Everything looks green except cloudflare publish which I think is unrelated to this PR:
> 
> `Error: Input required and not supplied: apiToken`

Yeah, that'll fail since you aren't in the org. Only outstanding thing to me is https://github.com/CrowdStrike/ember-toucan-core/pull/447#pullrequestreview-2465207019.  @simonihmig do you recall?

File: packages/ember-toucan-core/src/components/form/controls/textarea.gts [hunk 4]
```diff
@@ -80,7 +80,8 @@ export default class ToucanFormTextareaControlComponent extends Component<Toucan
         readonly={{@isReadOnly}}
         ...attributes
         {{on "input" this.handleInput}}
-      >{{@value}}</textarea>
```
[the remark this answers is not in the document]
simonihmig: Nope.

But quickly checking [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) shows this change here seems to be the correct thing to do:

image

simonihmig (approved):
Sorry it took me so long to get to this!

simonihmig:
Seems we have the same issue in headless-form [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/control/textarea.gts#L56). Created this ticket: https://github.com/CrowdStrike/ember-headless-form/issues/527

PR #466, opened 2024-11-27
Expose state from headless form

bradaune-cs:
## 🚀 Description

- [x] added: async validation and submission state pass-throughs from ember-headless-form
- [x] tested: new exposed async features
- [x] added: documentation for async state
- [x] added: documentation for yup validation

---

## 🔬 How to Test

Run the test suite to validate the new changes.

---

N/A

### DIFF
--- a/docs/toucan-form/async/demo/base-demo.md
+++ b/docs/toucan-form/async/demo/base-demo.md
@@ -1,0 +1,54 @@
+# Async state
+
+Submit this form with a valid email, and with the same email again, to see how it disables the submit button, changes its label, and shows error messages coming from the "backend":
+
+```hbs template
+<ToucanForm @onSubmit={{this.handleSubmit}} as |form|>
+  <form.Field @name='email' as |field|>
+    <div class='my-2 flex flex-col'>
+      <field.Label>Email</field.Label>
+      <field.Input
+        @type='email'
+        placeholder='Please enter your email'
+        class='border rounded px-2'
+      />
+    </div>
+  </form.Field>
+
+  <button type='submit' disabled={{form.submissionState.isPending}}>
+    {{if form.submissionState.isPending 'Submitting...' 'Submit'}}
+  </button>
+
+  {{#if form.submissionState.isResolved}}
+    <p>We got your data! 🎉</p>
+  {{else if form.submissionState.isRejected}}
+    <p>⛔️ {{form.submissionState.error}}</p>
+  {{/if}}
+</ToucanForm>
+```
+
+```js component
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class MyFormComponent extends Component {
+  saved = [];
+
+  @action
+  async handleSubmit({ email }) {
+    // pretending something async is happening here
+    await new Promise((r) => setTimeout(r, 3000));
+
+    if (!email) {
+      throw new Error('No email given');
+    }
+
+    if (this.saved.includes(email)) {
+      // Throwing this error will cause the form to yield form.submissionState.isRejected as true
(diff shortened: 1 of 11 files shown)


packages/ember-toucan-form/src/components/toucan-form.gts
```diff
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 import { hash } from '@ember/helper';
 
 import { HeadlessForm } from 'ember-headless-form';
+import { TrackedAsyncData } from 'ember-async-data';
```
simonihmig, line 5: This should have be a `type` import. Seems we don't have the linting setup that we have elsewhere to enforce this.

packages/ember-toucan-form/package.json
```diff
@@ -45,6 +45,7 @@
   "dependencies": {
     "@babel/runtime": "^7.20.7",
     "@embroider/addon-shim": "^1.0.0",
+    "ember-async-data": "^1.0.3",
```
simonihmig, line 48: I don't think this needs to be a `dependency`. We are not actively using it, only ember-headless-form does and that has it already as a dependency. With the import being changes to a type-only import (see below), this can be a devDependency I think!

simonihmig (commented):
Great stuff! 😍

Also thanks for the docs updates!!!

Just left some minor comments to look into!

packages/ember-toucan-form/package.json
bradaune-cs: done

docs/toucan-form/async/demo/base-demo.md
ynotdraw, line 1: Nice docs! Thank you. Fun example.

packages/ember-toucan-form/src/components/toucan-form.gts
```diff
@@ -14,6 +14,8 @@ import TextareaFieldComponent from '../-private/textarea-field';
 
 import type { HeadlessFormBlock, UserData } from '../-private/types';
 import type { WithBoundArgs } from '@glint/template';
+import type { TrackedAsyncData } from 'ember-async-data';
```
ynotdraw, line 17: Given we are using this package in client code, should the dependency be under dependencies rather than devDependencies?

(Fair warning: reviewing on mobile so I may be missing something 😅)

test-app/tests/integration/components/toucan-form/async-toucan-form-test.gts
```diff
@@ -0,0 +1,335 @@
...
+
+      await waitFor('[data-test-validation-state]');
+
+      assert
+        .dom('[data-test-validation-state]')
+        .hasText('PENDING', 'form.validationState is pending');
+      assert
+        .dom('[data-test-submission-state]')
+        .doesNotExist(
+          'form.submissionStatenis not present until validation has finished'
```
ynotdraw, line 312: Super nit:

```suggestion
          'form.submissionState is not present until validation has finished'
```

ynotdraw (approved):
Thanks a bunch @bradaune-cs!

packages/ember-toucan-form/src/components/toucan-form.gts
simonihmig: Brad changed this on my request actually! 😅 See https://github.com/CrowdStrike/ember-toucan-core/pull/466#discussion_r1860113683

simonihmig (approved):

packages/ember-toucan-form/src/components/toucan-form.gts
ynotdraw: Ah, makes sense. Thanks Simon! I missed that!

PR #642, opened 2025-05-30
fix: add `disabled` attribute to buttons

ryan-nauman-cs:
# 🚀 Description

This PR adds the native HTML `disabled` attribute to our Button component when `@isDisabled` is true. Previously, we were only setting `aria-disabled="true"` which provides accessibility information but doesn't actually prevent interaction with the button.

Changes include:
- [x] Added: Native `disabled` HTML attribute to the Button component
- [x] Updated: Tests to properly verify disabled button behavior
- [x] Fixed: Potential accessibility issue where disabled buttons could still be clicked programmatically

This change ensures that disabled buttons are truly non-interactive, improving both accessibility and user experience.

---

## 🔬 How to Test

1. Navigate to any page with buttons in the preview URL
2. Find or create a disabled button (using `@isDisabled={{true}}`)
3. Verify that:
   - The button appears visually disabled
   - The button cannot be clicked (cursor should show as not-allowed)
   - Screen readers announce the button as disabled
4. Try to programmatically click the button via browser console and verify it doesn't trigger the action

---

Before:
- Button appears disabled visually but could still be clicked programmatically
- Only had `aria-disabled="true"` attribute

image

After:
- Button has both `aria-disabled="true"` and native `disabled` attributes
- Button is completely non-interactive when disabled
- Tests properly verify the disabled state and behavior

image

### DIFF
--- a/packages/ember-toucan-core/src/components/button.gts
+++ b/packages/ember-toucan-core/src/components/button.gts
@@ -121,6 +121,7 @@ export default class Button extends Component<ButtonSignature> {
   <template>
     <button
       aria-disabled={{if @isDisabled "true"}}
+      disabled={{@isDisabled}}
       class={{this.styles}}
       type="button"
       {{on "click" this.onClick}}
--- a/test-app/tests/integration/components/button-test.gts
+++ b/test-app/tests/integration/components/button-test.gts
@@ -125,35 +125,67 @@ module('Integration | Component | button', function (hooks) {
     let handleClick = () => assert.step('clicked');
 
     await render(<template>
-      <Button @isDisabled={{true}} @onClick={{handleClick}} data-button>
-        button
-      </Button>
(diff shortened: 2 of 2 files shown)


mwagz (approved):

ynotdraw (commented):
Chatting about it internally, but I'm not sure this is what we want.

There is already https://github.com/ryan-nauman-cs/ember-toucan-core/blob/21946e78df33f2752e3d9b7372dbc06cb953007b/packages/ember-toucan-core/src/components/button.gts#L111-L116.

And adding the `disabled` attribute would take the component out of the tab order, meaning screen reader users won't be able to access our buttons. Normally disabled buttons are wrapped with a Tooltip, explaining why they are disabled. So that functionality would no longer be accessible.

ryan-nauman-cs:
discussed and agreed that we don't want to end up breaking things with this change.

PR #643, opened 2025-06-09
fix(multiselect): scope chevron clicks to the correct container

ryan-nauman-cs:
# Pull Request Template

## 🚀 Description

This PR fixes an issue with multiple multiselect components on the same page. Previously, when clicking on a multiselect container, the code would use `document.querySelector` to find the input element, which could incorrectly target the first multiselect on the page rather than the one that was clicked. 

The fix modifies the `handleContainerClick` method to:
- Accept a MouseEvent parameter
- Use the event's currentTarget to scope the querySelector to the specific container that was clicked
- This ensures each multiselect component correctly focuses its own input when clicked

Added a comprehensive test case that verifies multiple multiselect components work correctly when placed on the same page.

---

## 🔬 How to Test

1. Add multiple multiselect components to a page
2. Click on each multiselect container
3. Verify that the correct popover opens for the clicked multiselect
4. Verify that the input for the clicked multiselect receives focus
5. Run the new test case to confirm the fix works as expected

---

No visual changes to the component's appearance - this is a behavioral fix that ensures the correct multiselect responds when clicked. Here's a before & after of the docs that show the bad behavior.

https://github.com/user-attachments/assets/9d856d51-2292-463b-8e22-a4d18c5be1b2

https://github.com/user-attachments/assets/7822090b-e692-469a-815e-d9d84d4c0d40

### DIFF
--- a/.changeset/red-kiwis-eat.md
+++ b/.changeset/red-kiwis-eat.md
@@ -1,0 +1,5 @@
+---
+'@crowdstrike/ember-toucan-core': patch
+---
+
+Fixed multiselect container click handler to properly focus the correct input element when multiple components are present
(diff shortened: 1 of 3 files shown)


mwagz (approved):
Thanks!

simonihmig (approved):
LGTM! Thanks for the fix. But CI didn't run here, not sure why? 

- [x] Let's ensure we have a green build before merging!

PR #645, opened 2025-07-28
fix(autocomplete): assert nulls

ryan-nauman-cs:
## 🚀 Description

In development builds, deleting a value from autocomplete within a toucan form causes the value to be set to null. This causes the assertion to throw and the autocomplete to fail to render without a page refresh. This behavior does not occur in production builds.

---

## 🔬 How to Test

1. Use a development build
1. Render an autocomplete in a toucan form
2. Select a value
3. Delete the value
4. Collapse the autocomplete
5. Expand the autocomplete
6. Old behavior: empty container renders, New behavior: container renders as expected

---

https://github.com/user-attachments/assets/f497f665-5785-4c9e-95be-e0513a7c92ef

### DIFF
--- a/packages/ember-toucan-form/src/-private/autocomplete-field.gts
+++ b/packages/ember-toucan-form/src/-private/autocomplete-field.gts
@@ -51,10 +51,10 @@ export default class ToucanFormAutocompleteFieldComponent<
   @action
   assertSelected(value: unknown) {
     assert(
-      `A string or \`undefined\` is expected for ${String(
+      `A string or \`undefined\` or \`null\` is expected for ${String(
         this.args.name,
       )}, but you passed ${typeof value}`,
-      typeof value === 'string' || value === undefined,
+      typeof value === 'string' || value === undefined || value === null,
     );
 
     return value;
--- a/.changeset/breezy-rings-shave.md
+++ b/.changeset/breezy-rings-shave.md
@@ -1,0 +1,5 @@
+---
+'@crowdstrike/ember-toucan-form': patch
+---
+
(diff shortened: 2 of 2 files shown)


packages/ember-toucan-core/src/components/form/controls/autocomplete.gts
```diff
@@ -70,7 +70,7 @@ export interface ToucanFormAutocompleteControlComponentSignature {
     /**
      * The currently selected option.
      */
-    selected?: string;
+    selected?: string | null;
```
ryan-nauman-cs, line 73: i went back and forth on this change. i had to make it in order for linting to pass. an alternative option would be to trace back why null is being sent instead of undefined, however this felt like the riskier change to me as this behavior has been this way for some time. i wasn't comfortable with risking breaking existing consumer applications

packages/ember-toucan-form/src/-private/autocomplete-field.gts
```diff
@@ -51,10 +51,10 @@ export default class ToucanFormAutocompleteFieldComponent<
   @action
   assertSelected(value: unknown) {
     assert(
-      `A string or \`undefined\` is expected for ${String(
+      `A string or \`undefined\` or \`null\` is expected for ${String(
```
ynotdraw, line 54: Super nit, but could do:


```suggestion
      `A string, \`undefined\`, or \`null\` is expected for ${String(
```

ynotdraw (approved):

simonihmig (approved):

